PR #1, opened 2017-05-26
Add link to mailing list (google group)

duglin:


README.md
```diff
@@ -17,4 +17,8 @@ For more information, see this collaborative [document](https://docs.google.com/
   * Identifying areas where we’d like to see some harmonization or interop work
 * Create a white/position-paper on Serverless in relation to Cloud Native.
 
+## Communications
+
+We have google group for e-mail communications: [cncf-wg-serverless](https://groups.google.com/forum/#!forum/cncf-wg-serverless)
```
caniszczyk: do we really need a mailing list?

duglin: I assume we may have some out-of-band communications - or if nothing else to tell people when the meetings are :-)

luebken: FWIW I came here and was happy to find out that there is a mailing list

PR #6, opened 2017-06-27
Fix next meeting date (was Jul 5, should be Jul 6)

krook:


duglin:
LGTM I think this is just a typo from when the section was added

PR #17, opened 2017-10-25
Suggest using URNs for eventID and introduced eventGenerator

ieb:
I dont know if pull requests from the community are appropriate, but this small change might improve the scalability of the proposal. Please feel free to modify, reject (or ignore/delete) the pull request if not appropriate.

In large systems, non namespaced eventIds will require that events are wrapped in an envelope to avoid collisions. There are no guarantee that a producer won't use a counter for its event ID. The urn provides a standards based mechanism of making the ID unique while still allowing the producer the flexibility to use the a commit ID.

Similarly without some context provided by the eventGenerator the consumer will require some out of band knowledge of the meaning of the event. Having a URI (urn or stable url) as an eventGenerator provides a mechanism by which the consumer can discover the meaning of data elements, without having to infer context from the eventType. Abstract eventTypes like "document.updated", "asset.created" may have different significance in different contexts.

Consumers don't have to resolve the urns or get the stable urls if a simple string match is sufficient, but it gives some context without additional envelopes.

### DIFF
--- a/proposals/open-events/README.md
+++ b/proposals/open-events/README.md
@@ -14,7 +14,9 @@ Version 0.3 - 2017/09/20
 
 # Schema
 
-- **eventId** - *string* - ID of the event.  Can be specified by the producer.  The semantics of this string are explicitly undefined to ease the implementation of producers (e.g. a database commit ID).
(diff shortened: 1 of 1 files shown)


mrutkows:
I agree, as long as what you mean by URN is defined by the URN scheme [RFC2141] where it is a URI to guarantee global uniqueness.

duglin:
@ieb thanks for this. We've created a new repo for the eventing spec - can you open up this PR over there?  https://github.com/openeventing/spec/

Closing here since the work will continue over in the openeventing repo.

PR #33, opened 2018-02-14
Clean up README

caniszczyk:
Also we should version the whitepaper.

Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>


caniszczyk:
Pushing this in before the announcement at 830am PT.

PR #49, opened 2018-04-04
Fixed markdown formatting

mateimicu:
* Fix bold usage
* Update link to section


duglin:
LGTM  

thanks

Looks like there are other links that are bad too :-(     will try to fix those soon

PR #51, opened 2018-05-10
add work stream proposals

duglin:
Signed-off-by: Doug Davis <dug@us.ibm.com>


duglin:
From today's call (and f2f last week). Feel free to open a PR to add more details if needed.
We will be reviewing this list on next week's call and deciding what to work on next, and when.

PR #52, opened 2018-05-11
Details for some workstreams proposals

Vlaaaaaaad:
Hi,

Added some details and context for 2 of the workstreams proposals so the discussion on the next week's call will start on a stronger base.

The `README.md` is getting a bit large, but I assume the proposals will get their own folders after the next week's call( if they will be going forward).


proposals/README.md
```diff
@@ -23,8 +23,180 @@ on-going work for the current spec is hosted at
...
 
+The way functions are orchestrated, workflows are built and chaining functions were deemed to be not in scope for CloudEvents, but maybe in scope for the Serverless working group.
+
+There are issues such as the [correlation id](https://github.com/cloudevents/spec/pull/128), [event history/ chaining](https://github.com/cloudevents/spec/issues/204), [event nesting](https://github.com/cloudevents/spec/issues/72) and concerns regarding encryption/ signed content/ integrity that do need this to be defined. There are increasing number of question regarding who is allowed to modify a CloudEvent, how an event is forwarded and so on.
+
+Having a field in CloudEvents that is to be used for say correlation, but no clear specification for who uses that field and what that field contains is not ideal.
+
 ### Function Signatures
 
+There are multiple providers that have different ways to handle functions. Using multiple providers, switching providers and developing functions would be significantly better experiences if there was a common structure for function signatures.
```
duglin: lets keep a short description here but then move the other text/examples into a new md file and just reference it from here.  ok?

Vlaaaaaaad: Sure, I can do that later today. Should I create a new folder and put the file in there say `function-signatures/README.md` or just add a `function-signatures-examples.md` file?

duglin: a new sibling file is fine. I don't think we need too much structure in there, yet.  thanks

duglin:
merging to get the ideas out there - thanks

PR #79, opened 2019-09-19
Specification Update - 2

tsurdilo:
I changed the spec document quite a bit, will try to summarize changes:

* Moved use cases to its own markup file - linked from the spec document
* Moved examples to its own markup file - linked from the spec document
* Updated all image names in /media so they make sense
* Instead of a lot of repetitive and not so intuitive wording, defining the serverless workflow 
states and triggers via JSON Schema definitions. This is not only much easier to understand (each property includes a description field) but also makes it easier for implementors to understand what they are implementing actually)
* Adding the entire JSON Schema document - with comment that just like this doc it is a work in progress.
* Updating wording ..there was a lot of duplication and repetitiveness. Also same things were called with different names etc.
* Reduces size of images in spec document so its not 10 miles long.
* Updated contributor doc to add myself and Ruben (hope you guys don't mind)

Let me know what you think!


duglin:
@cathyhongzhang ?

cathyhongzhang:
I am reviewing it now.

Thanks,
Cathy

tsurdilo:
Hi @cathyhongzhang is this OK to be included? Any suggestions for changes on your end ? Thanks

cathyhongzhang:
Will post comments this afternoon. Thanks

workflow/spec/spec.md
```diff
@@ -1,311 +1,145 @@
-# Workflow - Version 0.1
+## Serverless Workflow Specification
 
-## Abstract
+As serverless applications are becoming more and more complex, there is need for
+coordination and orchestration of their operational flow. Workflows fit this need nicely,
+and many serverless workflow solutions exist today. The issue we see is that 
+many if not all current solutions have their own proprietary ways of describing serverless application flow and there
+currently no standardization between them.
```
cathyhongzhang: missing "is"

workflow/spec/spec.md
```diff
@@ -1,311 +1,145 @@
...
-  
-  -[Filter Mechanism](#Filter-Mechanism)
-  
-  -[Example](#Example)
-  
-  -[Errors](#Errors)
+- [Use Cases](#Use-Cases)
+- [Specication Details](#Specification-Details)
+- [Information Passing](#Information-Passing)
+- [Filter Mechanism](#Filter-Mechanism)
```
cathyhongzhang: It is nice to move the "Use Cases" to another file. But the "Information Passing" and "Filter Mechanism" should be part of the "Specification Details". Could you move them to be subsections of the "Specification Details"?

workflow/spec/spec.md
```diff
@@ -1,311 +1,145 @@
...
 ## Introduction
 
-Many serverless applications are not a simple function triggered by a single
-event, instead they are composed of multiple steps of function execution with
-functions in different steps triggered by different events. If a step involves
-multiple functions, the functions in that step might execute in sequence or in
-parallel or in branches depending on different event triggers. In order for a
-serverless platform to execute a serverless application's function workflow
-correctly, the application developer needs to provide a workflow specification.
-
```
cathyhongzhang: Did you move this section to another file or did you delete it? I would like to keep this section to clarify why we need a workflow specification.

workflow/spec/spec.md
```diff
@@ -1,311 +1,145 @@
...
-for the users to specify their serverless application workflow to help
-facilitate portability of the Serverless application across different vendors'
-platforms.
-
-This specification will meet the above goals as it describes a complete
-contract such that a given event timeline and workflow always produces the same
-set of side-effects.
-
-## Use Case Examples
-
```
cathyhongzhang: Could you confirm that what you did is just move this whole section to another file and you do not make any change on these examples?

workflow/spec/spec.md
```diff
@@ -1,311 +1,145 @@
...
+* **Orchestrate serverless application logic**: serverless applications are typicall event-driven and can be 
+very hard to manage. Serverless Workflow groups the application events and functions into a coherent unit and 
+simplifies orchestration of the app logic.
+* **Define and manage application data flow**: allows users to define how data is passed from incoming events to states triggered by those events
+as well as data passed between states themselves.
+* **Define and coordinate application operational state**: allow users to define the execution/operation
+flow and how/which functions are to be invoked on arrival of events.
+
+Following example illustrates a Serverless Workflow that involves events
+and functions. It specifies the interaction between events, states and functions to be invoked.
```
cathyhongzhang: Suggestion to change to: 
**Define and coordinate application control flow**: allow the users to define the execution/operation
control flow and how/which functions are to be invoked on arrival of events.
**Define and manage application data flow**: allows the users to define how data is passed and filtered from incoming events to states, from states to functions, from one function to another function, and from one state to another state.

workflow/spec/spec.md
```diff
@@ -1,311 +1,145 @@
...
+    "type": "object",
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Workflow name",
+            "minLength": 1
+        },
+        "exec-status": {
+            "type" : "string",
+            "enum": ["SYS.Timeout", "SYS.Fail", "SYS.Success", "SYS.MatchAny", "SYS.Permission", "SYS.InvalidParameter", "SYS.FilterError"],
```
cathyhongzhang: I am OK with adding the following generic status code: "SYS.Timeout", "SYS.Fail", "SYS.Success", "SYS.InvalidParameter". As to the others, they are not at the same logic level as these high-level system errors. People would ask why we have a "filter error" but not a "event parsing error"?

workflow/spec/spec.md
```diff
@@ -1,311 +1,145 @@
...
+            "description": "CloudEvent eventID"
+        },
+        "correlation-token": {
+            "type": "string",
+            "description": "Path string to an identity label field in the event message"
+        },
+        "message": {
+            "type": "string",
+            "description": "Trigger event message"
+        }
```
cathyhongzhang: What is this message field used for?

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
-with the workflow) is an array of event triggers associated with a Function
-Workflow. If there are multiple events involved in an application workflow, a
-correlation-token, which is used to correlate an event with other events for
-same workflow instance, must be specified in that event trigger.
+Note that each event definition has a "next-state" property, which is used to idetify the state which 
+should get executed after this event completes (value should be the unique name of a state).
 
-The states array (required) is an array of states associated with a Function
-Workflow.
+Each event state event can include one or more actions which can be executed, let's define those now:
```
cathyhongzhang: typo in this sentence.

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
 
-At high level, the Function Workflow specification consists of two parts:
-trigger definitions and state definitions.
+#### Event State
+* **Json Schema**
+```json
+{
+    "type": "object",
+    "description": "This state is used to wait for events from event sources and then to invoke one or more functions to run in sequence or in parallel.",
+    "extends": {
```
cathyhongzhang: "extend" seems to be a very old JSON schema keyword. Let's not use it.

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
+                "$ref": "#definitions/event"
+            }
+        }
+    },
+    "required": []
+}
+```
+Event state can hold one or more events definitions, so let's define those:
 
+#### Event State - event definitions
```
cathyhongzhang: The following is not event definition.  It is "event state" definition. There seems no need for this session title.

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
-trigger definitions and state definitions.
+#### Event State
+* **Json Schema**
+```json
+{
+    "type": "object",
+    "description": "This state is used to wait for events from event sources and then to invoke one or more functions to run in sequence or in parallel.",
+    "extends": {
+        "$ref": "#definitions/defaultstate"
+    },
```
cathyhongzhang: Why do we need this properties part? We do not need to define events inside a event state. Event is defined in trigger definition, which is outside of the event state.

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
+Event state can hold one or more events definitions, so let's define those:
 
+#### Event State - event definitions
+* **Json Schema**
 ```json
 {
-  "trigger-defs" : [],
-  "states": []
+    "type": "object",
+    "description": "Event associated with a State",
```
cathyhongzhang: description should be "This state is used to wait for events from event sources and then to invoke one or more functions to run in sequence or in parallel." . It is better to redo and simplify the event state scheme.

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
+                "$ref": "#definitions/action"
+            }
+        },
+        "next-state": {
+            "type": "string",
+            "description": "Name of the next state to transition to after all the actions for the matching event have been successfully executed",
+            "minLength": 1
+        }
+    },
+    "required": []
```
cathyhongzhang: event expression, timeout, and next state should be "required" keys

workflow/spec/spec.md
```diff
@@ -1,311 +1,145 @@
...
+        "correlation-token": {
+            "type": "string",
+            "description": "Path string to an identity label field in the event message"
+        },
+        "message": {
+            "type": "string",
+            "description": "Trigger event message"
+        }
+    },
+    "required": ["name"]
```
cathyhongzhang: event source and event ID should be "required" since event "source ID +event ID" will uniquely identify an event. We need this unique ID in the event state "event expression" to identify the incoming event.

workflow/spec/spec.md
```diff
@@ -1,311 +1,145 @@
...
-    payload as it progresses through the Function Workflow.
-
-- Function Workflow provides a way for the application developer to specify a
-    unique field in the event that can be used to correlate events from the
-    event sources to the same function workflow instance
-
-A Function Workflow can be naturally modeled as a state machine. The following
-is a list of states provided by the definition/specification of a Function
-Workflow. The specification of a workflow is called a workflow template. The
-instantiation of the workflow template is called a workflow instance.
```
cathyhongzhang: Let's keep the functional scope and workflow model since these were added to clarify the questions people raised before. If needed, we can create another PR to simplify it.

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
+    "required": []
 }
 ```
+The event expression attribute is used to associate this event state with one or more trigger events. 
 
-The trigger-defs array (only required if there are events associated
-with the workflow) is an array of event triggers associated with a Function
-Workflow. If there are multiple events involved in an application workflow, a
-correlation-token, which is used to correlate an event with other events for
-same workflow instance, must be specified in that event trigger.
```
cathyhongzhang: This part should not be removed. It should be moved to the trigger-def section.

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
+    "properties": {
+        "function": {
+            "$ref": "#definitions/function",
+            "description": "Specifies the function that must be invoked"
+        },
+        "timeout": {
+            "type": "integer",
+            "default":"0",
+            "minimum": 0,
+            "description": "Specifies the time period waiting for the events in the EVENTS-EXPRESSION"
```
cathyhongzhang: This timeout description is not correct. It should be "field specifies the maximum amount of time in seconds to wait for the completion of the function's execution. This must be a positive integer. The function timer is started when the request is sent to the invoked function"

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
+{
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "Function name",
+      "minLength": 1
+    },
+    "type": {
+      "type": "string",
```
cathyhongzhang: Could you clarify what type is meant here?

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
-   ]
+{
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "Function name",
+      "minLength": 1
+    },
+    "type": {
```
cathyhongzhang: Could you clarify what type is meant here?

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
+        "next-state": {
+            "type": "string",
+            "description": "Name of the next state to transition to when exceeding max-retry limit",
+            "minLength": 1
+        }
+    },
+    "required": []
 }
 ```
+Results from a serverless function can be used to initiate a retry which can determine if another function(s)
```
cathyhongzhang: I don't think we need to add this. It adds more confusion.

workflow/spec/spec.md
```diff
@@ -319,452 +153,526 @@ instantiation of the workflow template is called a workflow instance.
...
+            }
+        },
+        "next-state": {
+            "type": "string",
+            "description": "Name of the next state to transition to after all the actions have been successfully executed",
+            "minLength": 1
+        }
+    },
+    "required": []
+}
```
cathyhongzhang: need to add required keys

workflow/spec/spec.md
tsurdilo: I replaced it with the first section of the document in pr:
"## Serverless Workflow Specification"

I thought that this section tries to explain what an event-based application is and what a workflow is, but readers interested in this spec already know that. Rather than trying to give some explanation, I focused the to explain why a workflow is useful for event based apps, by using words like "coordination" and "orchestration". Let me know if that's OK or if you definitely want to put this back.

workflow/spec/spec.md
tsurdilo: The use cases were not deleted. They are just in a separate file now -- spec-usecases.md

workflow/spec/spec.md
tsurdilo: It is the payload of the incoming event. Would you rename it to something else?

workflow/spec/spec.md
tsurdilo: Good point, I have removed "extends" from all places in the document.

workflow/spec/spec.md
tsurdilo: The event state defines an array of events:
```json
"events": {
            "type": "array",
            "description": "Event State Definitions",
            "items": {
                "type": "object",
                "$ref": "#definitions/event"
            }
        }
```
What the mentioned section describes is the definition of the event array of the event state. Yes a little confusing I know :) So I think this section is a valid one, but let me know suggestions of renaming it if you feel would cause confusion for readers.

workflow/spec/spec.md
tsurdilo: These are different kinds of events. "trigger-defs" defines the incoming or triggering events. An Event State defines an event array which are events that happen when this state is active. 
This is just carried over from the currently active document (see "Event State" section in here https://github.com/cncf/wg-serverless/blob/master/workflow/spec/spec.md). 

Now that I see this maybe it's best to rename the "events" array in the Event State to maybe "steps" or "activities". WDYT?

workflow/spec/spec.md
tsurdilo: This goes back to previous comment - this is not a definition of the Event State (which is one section above). It is the definition of "events that happen when the Event State is active".

I agree completely with renaming this to reduce confusion, as mentioned before to maybe "steps" or "activities". Let me know what you think.

workflow/spec/spec.md
tsurdilo: good point! added them back, but did a little rewording

workflow/spec/spec.md
tsurdilo: added back under the "Trigger Definition" section

workflow/spec/spec.md
tsurdilo: It is a string field that allows implementors to define a type of function (for example in aws you can define AWS::Serverless::Function, AWS::Serverless::Applicatio, etc etc). I thought this would good addition so implementors would know how to implement the function call better. It is not a required field.

workflow/spec/spec.md

workflow/spec/spec.md

tsurdilo:
@cathyhongzhang Thanks for the great review. I have updated the PR with changes and there are some comments/info/questions about some of your suggestions. Please look at those when you have the time. Thanks!

workflow/spec/spec.md
ruromero: Question. What would you put in the eventID? This should be unique to each event. Same applies to message. I would however add `type`. Below an example of how I think an event trigger would be defined:
```
    {
      "name": "car.shipped",
      "source": "http://workflows.example.com/factory/",
      "type": "com.myspace.carfactory.shipment.done",
      "correlation-token": "$.vin"
    }
```

workflow/spec/spec.md
```diff
@@ -319,452 +191,548 @@ instantiation of the workflow template is called a workflow instance.
...
+        }
+    },
+    "required": ["name", "type"]
 }
 ```
 
-- The **"status"** field specifies the workflow termination status: SUCCESS or
-FAILURE.
+Parallel state defines a collection of branches which are to be executed in parallel. 
+Each branch consists of a collection of states to be executed. Let's define a branch now:
```
ruromero: I would add to the spec whether a branch must be completed before the sequence can continue. Maybe a `wait-for-completion` flag. In some cases the user would want to execute all branches in parallel and wait until all are finished or maybe it is not needed and all must be executed but once the first branch is done, the workflow execution can continue. WDYT?

tsurdilo: very good idea. i will add wait-for-completion boolean param to parallel state. the default should be "false" imo and be as you mentioned that flow can continue as soon as first branch completes. if this flag is set to true then all branches must complete before flow can continue. Thanks!

tsurdilo: updates - let me know if it's OK now.

workflow/spec/spec.md
tsurdilo: @ruromero good point! - updated the definition
@cathyhongzhang good point as well about message - i removed it.

updated

workflow/spec/spec.md
cathyhongzhang: This section is trying to explain why we need a workflow spec for serverless. You and I are convinced that we need it, but some people would think a lambda-like single-event-triggered function is enough. I think it is good to include this in the introduction.

workflow/spec/spec.md
cathyhongzhang: I assume that you have not uploaded the updated version yet, right?

workflow/spec/spec.md
cathyhongzhang: CLoudEvents spec already defined the "type" field as well as other fields associated with an event. W.r.t. an event, "source"+"eventID" uniquely identifies an event. So in the workflow, we only need source and eventID to uniquely identify an event. The type does not have any effect on an workflow logic. So I do not think we need it.

workflow/spec/spec.md
cathyhongzhang: This does not look like an event definition. It is more about the interaction between events and actions, which should be part of the event state definition. We do not need to make this so complicated. We just need to define the event state here which includes the event expression, the actions, next state.  As to detail event definition, it is covered by CloudEVents spec

workflow/spec/spec.md
tsurdilo: @cathyhongzhang good point - I have added this section back, but reworded it so it sounds better. Hope its ok now.

workflow/spec/spec.md
tsurdilo: @cathyhongzhang its there, look at https://github.com/cncf/wg-serverless/pull/79/files#diff-bc18ddd43c9fef122edf80ec220f04bbR38

workflow/spec/spec.md
tsurdilo: @cathyhongzhang ok, updated to reflect your comments.

tsurdilo: @ruromero i found the type attribute under "required attributes" -- https://github.com/cloudevents/spec/blob/master/spec.md

workflow/spec/spec.md
tsurdilo: @cathyhongzhang i agree with that, but do you mind if I work on that in the next PR?
This one has become very large, and maybe we can focus on reducing the complexity one state at a time in smaller prs? wdyt?

tsurdilo:
@cathyhongzhang thanks for additional comments. I have made updates based on them. As far as reducing complexity of some states, would you mind if we do that in smaller future prs as this one has become quite large.

workflow/spec/spec.md
ruromero: @cathyhongzhang @tsurdilo Yes, I know `type` is mandatory so is the ID. My point is why you need to add it to the trigger-def event definition.
The eventID is unique so what would a user define in the eventID and how would be used in the event-expression?
That's why I see the type useful. In the example above the user would be able to filter by name/source/type or even if the correlation-token is empty.
```
event-expression: name == 'car.shipped'
event-expression: !correlation-token
event-expression: type == 'com.myspace.carfactory.shipment.done' or type == 'com.myspace.carfactory.shipment.cancelled'
```

Am I misunderstanding something?

workflow/spec/spec.md
cathyhongzhang: I can not find your update in this PR. Is your update in another PR that you have not uploaded yet?

workflow/spec/spec.md
cathyhongzhang: got it. Thanks

workflow/spec/spec.md
cathyhongzhang: I did not see the following being added back :  
1. Define the steps/states and workflow involved in a serverless application.
2. Define which functions are involved in each step.
3. Define which event or combination of events trigger the function or
    functions.
4. Define how to arrange those functions to execute in sequence or in parallel
    if multiple functions are triggered.
5. Specify how information is filtered and passed from the event to the
    function or between functions or between states.
6. Define error conditions with retries.
7. If a function is triggered by two or more events, define what label/key
    should be used to correlate those events to the same function workflow
    instance.

workflow/spec/spec.md
tsurdilo: @cathyhongzhang its right on top in the "Abstract" section:
```
Serverless applications are becoming increasingly complex. Nowdays they have to coordinate, manage, and define
the execution order (steps) for countless functions triggered by as many events.

When we are dealing with large number of functions, managing their execution is not a simpler task. 
For example we have to coordinate functions and event triggers, orchestrate function execution (sequential,  parallel, 
in branches depending on different event triggers), etc.

Workflows have become key components of serverless applications as they excel at orchestration and coordination
of their functional flow. 

The goal of the Serverless Workflow sub-group is to come up with a standard way for users to specify their serverless application workflow, as well as help facilitate 
portability of serverless applications across different vendor platforms.

Serverless Workflow is a vendor-neutral and portable specification which meets these goals.
```

workflow/spec/spec.md
```diff
@@ -1,9 +1,21 @@
-# Workflow - Version 0.1
+# Serverless Workflow - Version 0.1
 
 ## Abstract
 
-Workflow is a vendor-neutral specification for defining the format/primitives
-that the users can use to specify/describe their serverless application flow.
+Serverless applications are becoming increasingly complex. Nowdays they have to coordinate, manage, and define
+the execution order (steps) for countless functions triggered by as many events.
```
cathyhongzhang: remove "as"?

workflow/spec/spec.md
```diff
@@ -1,9 +1,21 @@
...
+
+When we are dealing with large number of functions, managing their execution is not a simpler task.
```
cathyhongzhang: change to simple

workflow/spec/spec.md
cathyhongzhang: OK, Thanks.

workflow/spec/spec.md
cathyhongzhang: The eventID and event source defined here in the workflow spec are used to uniquely identify an event. They are used to refer to the event object as defined according to the CloudEvents spec or to any other proprietary event spec if an implementation likes to use a proprietary event spec. We need these two fields to uniquely identify one event from another event. Maybe it is good to add some clarification on this in this spec. The event expression defined in the event state will be an "and"/"or"/"not" operation of these event IDs.

workflow/spec/spec.md
cathyhongzhang: yes, I prefer smaller PRs. But this is an important point--- how we define and structure the event state and event. I would like you to change this according to our agreement before it is merged. Thank you!

workflow/spec/spec.md
cathyhongzhang: Agree. Could you add some explanation here?

workflow/spec/spec.md
cathyhongzhang: I think we need to revisit the "required" field at each place later. Let's create another separate PR to sync up on the "required".

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+    "description": "Permits transitions to other states based on criteria matching.",
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Unique name of the state"
+        },
+        "type": {
+            "type" : "string",
+            "enum": ["EVENT", "OPERATION", "SWITCH", "DELAY", "PARALLEL", "END"],
+            "description": "State type"
```
cathyhongzhang: type can only be SWITCH. Can not be other string.

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+    "type": "object",
+    "description": "This state allows one or more functions to run in sequence or in parallel without waiting for any event.",
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Unique name of the state"
+        },
+        "type": {
+            "type" : "string",
+            "enum": ["EVENT", "OPERATION", "SWITCH", "DELAY", "PARALLEL", "END"],
```
cathyhongzhang: can only be "EVENT".

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+        },
+        "choices": {
+            "type": "array",
+            "description": "Defines an ordered set of Match Rules against the input data to this state",
+            "items": {
+                "type": "object",
+                "anyOf": [
+                    { "$ref": "#definitions/andchoice" },
+                    { "$ref": "#definitions/notchoice" },
+                    { "$ref": "#definitions/orchoice" }
```
cathyhongzhang: We should allow a matching rule without any of and/or/not. This syntax does not seem to allow this.

cathyhongzhang: Where is the following fields defined? 
               "path":"PAYLOAD-PATH",
               "value":"VALUE",
               "operator":"COMPARISON-OPERATOR",

We need define the path filter, value to be matched, and the operator. 

Also it misses the "next state" field.

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+        },
+        "value": {
+            "type": "string",
+            "description": "Matching value"
+        },
+        "operator": {
+            "type" : "string",
+            "enum": ["EQ", "LT", "LTEQ", "GT", "GTEQ", "StrEQ", "StrLT", "StrLTEQ", "StrGT", "StrGTEQ"],
+            "description": "Specifies how the input data is compared with the values",
+            "format": "validoperator"
```
cathyhongzhang: Why do we need this format?

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
-workflow will transit to end state.
-- The **"action-mode"** field specifies whether functions are executed in sequence or
-in parallel and could either be SEQUENTIAL or PARALLEL.
-- The **"actions"** field is a list of
-[Action definitions](#action-definitions) constructs that
-specify the list of functions to be performed when the event that matches the
-event-expression is received.
-- The **"next-state"** field specifies the name of the next state to transition to
-after all the actions for the matching event have been successfully executed.
-
```
cathyhongzhang: This event state needs to be added back!

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+{
+    "type": "object",
+    "description": "And Choice",
+    "properties": {
+        "and": {
+            "type": "array",
+            "description": "List of choices",
+            "items": {
+                "type": "object",
+                "$ref": "#definitions/defaultchoice"
```
cathyhongzhang: The defaultchoice has "next state". But this operation should operate on two objects without "next state", i.e. "next state" should be put outside of the items.

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+{
+    "type": "object",
+    "description": "And Choice",
+    "properties": {
+        "or": {
+            "type": "array",
+            "description": "List of choices",
+            "items": {
+                "type": "object",
+                "$ref": "#definitions/defaultchoice"
```
cathyhongzhang: The defaultchoice has "next state". But this operation should operate on two objects without "next state", i.e. "next state" should be put outside of the items.

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+        "operator": {
+            "type" : "string",
+            "enum": ["EQ", "LT", "LTEQ", "GT", "GTEQ", "StrEQ", "StrLT", "StrLTEQ", "StrGT", "StrGTEQ"],
+            "description": "Specifies how the input data is compared with the values",
+            "format": "validoperator"
+        },
+        "next-state": {
+            "type": "string",
+            "description": "Specifies the name of the next state to transition to if there is a value match"
+        }
```
cathyhongzhang: "next state" should be removed. See my other comments below.

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+    "type": "object",
+    "description": "Causes the workflow execution to delay for a specified duration",
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Unique name of the state"
+        },
+        "type": {
+            "type" : "string",
+            "enum": ["EVENT", "OPERATION", "SWITCH", "DELAY", "PARALLEL", "END"],
```
cathyhongzhang: type can only be DELAY

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+    "type": "object",
+    "description": "End State",
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Unique name of the state"
+        },
+        "type": {
+            "type" : "string",
+            "enum": ["EVENT", "OPERATION", "SWITCH", "DELAY", "PARALLEL", "END"],
```
cathyhongzhang: ditto

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+        "start": {
+            "type": "boolean",
+            "default": false,
+            "description": "Is the start event"
+        },
+        "status": {
+            "type" : "string",
+            "enum": ["SUCCESS", "FAILURE"],
+            "description": "Specifies the workflow termination status",
+            "format": "validstatus"
```
cathyhongzhang: ditto

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+            "description": "Unique name of the state"
+        },
+        "type": {
+            "type" : "string",
+            "enum": ["EVENT", "OPERATION", "SWITCH", "DELAY", "PARALLEL", "END"],
+            "description": "State type"
+        },
+        "start": {
+            "type": "boolean",
+            "default": false,
```
cathyhongzhang: why do we need "start"?

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+    "type": "object",
+    "description": "Consists of a number of states that are executed in parallel",
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Unique name of the state"
+        },
+        "type": {
+            "type" : "string",
+            "enum": ["EVENT", "OPERATION", "SWITCH", "DELAY", "PARALLEL", "END"],
```
cathyhongzhang: ditto

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+                            { "$ref": "#definitions/parallelstate" },
+                            { "$ref": "#definitions/switchstate" }
+                        ]
+                    }
+        },
+        "wait-for-completion": {
+            "type": "boolean",
+            "default": false,
+            "description": "Flow must wait for this branch to finish before continuing"
+        }
```
cathyhongzhang: So we need to be clear on what the false behavior is. A branch could have a sequence of states, not just an invocation of one function. Suppose a workflow has two branches whose "wait-for-completion" are all set to false. The workflow invokes both branches which start execution on their first state. Then the main workflow will continue to go to next state while the two branch workflows continue their operations. But the spec says that end state is not allowed in branches, how does that branch flow end?

workflow/spec/spec.md
```diff
@@ -319,452 +193,542 @@ instantiation of the workflow template is called a workflow instance.
...
+    },
+    "required": ["name", "states", "wait-for-completion"]
 }
 ```
+Note that end states should not be allowed to be transitioned to. Also when executed the states triggered
+by parallel state should be treated as "single states", their next-state property should not be enabled.
```
cathyhongzhang: This statement is kind of confusing. Inside a branch which could have a sequence of states, next state should be allowed.

cathyhongzhang: In the original spec, a branch must have one state which does not have "next state" which denotes the end state of that branch. Now I am thinking maybe we should require each branch to have an explicit end state which denotes the end of that branch. If wait-for-completion is set, then the main flow should wait for that branch flow to reach end state. If it not set, then the main flow does not need to wait for that branch, and that branch will end execution separately when reaching end state. What do you think?

cathyhongzhang: Suggest to add the following back for clarification: 

Each branch receives a copy of the Parallel state's input data. 
"next-state" transitions for states within a branch can only be to other states in that branch. In addition, states outside a Parallel state cannot transition to a state within a branch of a Parallel state. 
The Parallel state generates an output array in which each element is the output for a branch. The elements of the output array need not be of the same type.

workflow/spec/spec.md
```diff
@@ -805,8 +773,7 @@ There are three kinds of filters
 - State Filter
   - Invoked when data is passed from the previous state to the current state
   - Invoked when data is passed from the current state to the next state
```
cathyhongzhang: Why do we delete the state filter?

workflow/spec/spec.md
```diff
@@ -823,83 +790,15 @@ Each Filter has three kinds of path filters
   - Specify output data of State or Action as JSONPath
   - Default value is '\$'
```
cathyhongzhang: Why do we delete these explanation for action filter?

workflow/spec/spec.md
```diff
@@ -823,83 +790,15 @@ Each Filter has three kinds of path filters
...
 
-![Sequential filter diagram](media/2f84c6c905d1b66fc780f7ca6a444435.png)
-
-![Parallel filter diagram](media/466d322720c67a60013f516f880c302f.png)
-
```
cathyhongzhang: Why do we delete these diagrams?

workflow/spec/spec.md
```diff
@@ -823,83 +790,15 @@ Each Filter has three kinds of path filters
...
-         ],
-         "next-state":"STATE_END"
-      },
-      {  
-         "name":"STATE-END",
-         "type":"END"
-      }
-   ]
-}
-```
```
cathyhongzhang: We need these examples, which were added per others' request before. We can correct the syntax as needed

workflow/spec/spec.md
tsurdilo: @cathyhongzhang sure but can you explain what the agreement is please? If you look at current document: https://github.com/cncf/wg-serverless/blob/master/workflow/spec/spec.md

you see event state defined as:
```json
{  
         "name":"STATE-NAME",
         "type":"EVENT",
         "start":true,
         "events":[  
            {  
               "event-expression":"EVENTS-EXPRESSION",
               "timeout":"TIMEOUT-VALUE",
               "action-mode":"ACTION-MODE",
               "actions":[  
               ],
               "next-state":"STATE-NAME"
            }
         ]
      }
```

and actions array under "Actions Definitions:
```json
{  
   "actions":[  
      {  
         "function":"FUNCTION-NAME",
         "timeout":"TIMEOUT-VALUE",
         "retry":[  
            {  
               "match":"RESULT-VALUE",
               "retry-interval":"INTERVAL-VALUE",
               "max-retry":"MAX-RETRY",
               "next-state":"STATE-NAME"
            }
         ]
      }
   ]
}
```

This is the exact same this as we have here but just in json schema format. If you can show me what you would like the json to look like for event states that is different than what's in the current document I'd be happy to change but I don't fully understand what your ideas are. Please let me know.

workflow/spec/spec.md
tsurdilo: @cathyhongzhang there have not been removed but are in the spec-examples.md now with the intention to add more examples there in future prs. here is link from this pr to that file:

https://github.com/cncf/wg-serverless/pull/79/files#diff-9a7b704f28b4efabde79725a4c380f3cR1

workflow/spec/spec.md
tsurdilo: @cathyhongzhang sorry wdym that section is still there. One thing that the current document does not do is clearly define each property of each trigger/state. With these changes in place we can then go through each node and look at each property and make updates. 

For example this while filtering section imo should be revised and properties updated accordingly. Things like "Each Filter has three kinds of path filters" and State filters, nothing is shown in the current document as far as what properties it should translate to and it's imo something we should look at at some point.

workflow/spec/spec.md
tsurdilo: that whole section is still there. what was deleted?

workflow/spec/spec.md
tsurdilo: added.

workflow/spec/spec.md
tsurdilo: I don't think we are defining implementation details here as that is up to the implementors to determine how to continue the async branch execution if no wait-for-completion is set.

workflow/spec/spec.md
tsurdilo: no "start" property needed for endstate..removed

workflow/spec/spec.md
tsurdilo: updated all type enums in pr

workflow/spec/spec.md
tsurdilo: updated all type enums in pr

workflow/spec/spec.md
tsurdilo: updated all type enums in pr

workflow/spec/spec.md
tsurdilo: updated all type enums in pr

workflow/spec/spec.md
tsurdilo: not needed. removed all "format" references.

workflow/spec/spec.md
tsurdilo: removed

workflow/spec/spec.md
tsurdilo: good point. according to other related comment for this -- removed next-state from default choice

workflow/spec/spec.md
tsurdilo: its in the "Functional Scope" section

workflow/spec/spec.md
tsurdilo: added

workflow/spec/spec.md
tsurdilo: good point - updated all enum fields

workflow/spec/spec.md
tsurdilo: ok let's do that. Once we add this big pr I will start creating a bunch of small ones that will be much easier to review.

workflow/spec/spec.md

workflow/spec/spec.md
tsurdilo: ok good point -- added default choice as well

workflow/spec/spec.md
tsurdilo: it is there, just the description of fields as they are in json schema now under "description" property

workflow/spec/spec.md
tsurdilo: duplicate comment - this is resolved

tsurdilo:
@cathyhongzhang since this pr getting harder to review with so many updates I did want to show you that you can see the latest document formatted in my branch from which I do pr from -- https://github.com/tsurdilo/wg-serverless/blob/specupdate2/workflow/spec/spec.md). 
If that is easier for you to see the document as it would look when we merge this pr. Hope this helps some.

workflow/spec/spec.md
ruromero: This is clear for me. Let me explain it with different words.
In the trigger-defs a user will define what **types** of events might "trigger" som action in the workflow. That is why adding the eventID to the definition of the type doesn't make sense.
Using user stories:
* As a user I want all the events of source `mysource1` to start the workflow in state A
* As a user I want all the events of source `mysource2` to start the workflow in state B

I don't see why would make sense the following user story:
* As a user I want the event of source `mysource` and eventID `111-aaa-bbb` to start the workflow in state A

It is evident that a cloud event MUST have an eventID/type/source/specversion, otherwise it wouldn't be a valid cloudEvent. However, the eventID can't be used to define a trigger.
Do you see my point?
Sorry for this long discussion but I think it is very important for the spec.

tsurdilo:
Updated document to add tables with parameter name and description. There is also a link (expandable) not which shows the JSON definition for each element. Inclusion of table I think is easier for non-technical people, but technical people still have chance to see the json schema for everything.

workflow/spec/spec.md
tsurdilo: the "as many" refers to the "countless" so i think it is ok the way it is . is that ok ?

tsurdilo:
@cathyhongzhang hello, could we come to a resolution of this PR if possible? IMO it includes a ton of improvements which we should add and would allow us to then open new smaller pull requests where we can focus on specific improvements. 

Again, you can see what the document will look like with these changes here: https://github.com/tsurdilo/wg-serverless/blob/specupdate2/workflow/spec/spec.md

workflow/spec/spec.md
cathyhongzhang: I see your point. Thanks.

workflow/spec/spec.md
cathyhongzhang: OK, thanks.

workflow/spec/spec.md
cathyhongzhang: OK, thanks

workflow/spec/spec.md
cathyhongzhang: I think the syntax should add "defaultchoice" to the "anyof"?

workflow/spec/spec.md
cathyhongzhang: thanks

workflow/spec/spec.md
cathyhongzhang: Let's keep it as is now. We can upload new PR if needed.

workflow/spec/spec.md
cathyhongzhang: I feel more explanation on some fields will be better for the user. But we can have new PR for that.

workflow/spec/spec.md
cathyhongzhang: ok

workflow/spec/spec.md
cathyhongzhang: ok

workflow/spec/spec.md
cathyhongzhang: ok

workflow/spec/spec.md
cathyhongzhang: I just noticed that the syntax misses the overall final "next state" for the "parallel state". Could you add that in?

cathyhongzhang: The "wait-for-completion" should be moved into each branch since my understanding is that it is associated with a branch, i.e. the parallel state can choose to wait for completion of some branches while not waiting for completion of other branches

cathyhongzhang: If we want to add " wait-for-completion", we should add the following rule to avoid ambiguous ending status of branches in the case of wait-for-completion being false:
1. If a branch's wait-for-completion is false, that branch must have an end state defined.
2. If a branch's wait-for-completion is true,  that branch should not have an end state and should have at least one state that has no next state.

workflow/spec/spec.md
cathyhongzhang: ok

workflow/spec/spec.md
cathyhongzhang: OK, the diff is a little confusing:-)

workflow/spec/spec.md
cathyhongzhang: OK

workflow/spec/spec.md

workflow/spec/spec.md
tsurdilo: sure. that will be fine

workflow/spec/spec.md
tsurdilo: @cathyhongzhang parallel state does have the ""next-state" property. it is there.

tsurdilo: @cathyhongzhang "wait-for-completion" parameter is in the branch definition and not the parallel state itself already.

tsurdilo: @cathyhongzhang I find that the two rules you mention are really not needed, especially since for rule 2 an end state is a state and does not have "next-state" property..so it's not really needed. I think describing wait-for-completion as it is now is enough for implementors to understand the execution behavior enough. We should really not do so many details as implementations can be very different and should allow implementors to do how they like without giving them so many details.

workflow/spec/spec.md
tsurdilo: yeah i agree. its gotten very large. but if we push it we will create smaller ones from now on :)

tsurdilo:
@cathyhongzhang @ruromero 
I made update after the latest comments from @cathyhongzhang (adding back type to trigger defs per your conversation). 
Can you guys look over the doc again (much easier just look at https://github.com/tsurdilo/wg-serverless/blob/specupdate2/workflow/spec/spec.md
and then lets please make a decision to include this pr or not finally :) 

Thanks a lot.

workflow/spec/spec.md
tsurdilo: next-state is present already
wait-for-completion is already in branch definition
i don't think these two rules are necessary. we should not give implementors so many details on how they want to implement parallel state i think.

tsurdilo: end state is a state and does not have "next-state" property..so i dont think that rule 2) would apply even. I really think we should allow implementors to do this as they wish and try to avoid putting in many implementation details as possible.

ruromero:
@tsurdilo I think it is much better now. In a follow up PR I would try to make navigation easier with anchors and links. You know in large documents with lots of definitions users like to easily jump to one section to another.
Otherwise it looks good to me. Great work.

tsurdilo:
@ruromero thanks. I added a bunch of anchor links, for states and also inside each table if a parameter has its own definition it is linked now. Hope this helps some. We can improve this in future.

workflow/spec/spec.md
cathyhongzhang: Have you added the "single choice" per my above comment?

workflow/spec/spec.md
cathyhongzhang: I understand your point. The two rules are added to be consistent with the following statement that describes the parallel state: 
"Each branch consists of a collection of states to be executed and continues its execution until it reaches a state that has no next state within the branch." 
If a branch has an end state, it seems to conflict with the above statement about reaching a state that has no next state within the branch. 
An alternative is to remove this statement and requires that each branch has an end state which specifies the completion of execution of that branch (it does not matter whether wait-for-completion is true or false). What do you think?

workflow/spec/spec.md
cathyhongzhang: Could you take out the other  status codes? We can add more in separate PR which explains what each new status code mean.

workflow/spec/spec.md
cathyhongzhang: Yes, it is better to rename it. We can take care of this in a separate PR.

cathyhongzhang:
I just replied to your replies:-) The key issue is the parallel state. After that is taken care of, I think we are good to go. Thanks!

workflow/spec/spec.md
tsurdilo: @cathyhongzhang yes I did, only ones specified are per your instructions:
"enum": ["SYS.Timeout", "SYS.Fail", "SYS.Success", "SYS.InvalidParameter"],

workflow/spec/spec.md
tsurdilo: @cathyhongzhang ok, I'll do one after this one is merged

workflow/spec/spec.md
tsurdilo: @cathyhongzhang I have updated the text to say that there should be an end state present to represent the completion of the branch. The issue I think we should address that spec says each state can possibly be a start state (with the "start" property") but then we have a specific end state. 
IMO we should in new PR think about removing end state and allow any state to be start or end. I think this is more flexible. We could then move the "status" property of end state currently to the workflow definition itself. wdyt?

workflow/spec/spec.md
tsurdilo: @cathyhongzhang added now - called "singlechoice"

tsurdilo:
@cathyhongzhang thanks for additional comments. I have updated the doc based on them.

workflow/spec/spec.md
cathyhongzhang: OK. Looks good to me now. The end state explicitly denotes ending of a branch or a workflow. The state preceding the end state can be any state. If we do not have end state, we have to add a flag to all the other states to specify whether it is an ending state or not.

cathyhongzhang:
The latest update looks good to me. It is ready to be merged. Thanks for this effort!

cathyhongzhang:
> @cathyhongzhang ?

LGTM.

PR #82, opened 2019-10-21
Removing end state

tsurdilo:
There is no need for us to have an explicit end state. in this pr:
* Adding "starts-at" property to workflow and branch definitions. This property defines the name of the state which is the starting state of the workflow.
* Adding "end" property to each state which states if this state is the ending state of the workflow.
* Removes end state
* Updates text and small fixes

I think this is a much better option and will:
* make the overall json shorter
* allow more flexibility for implementations
* follows closer other workflow impls such as step functions for example

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -9,10 +9,14 @@
       "description": "Workflow name",
       "minLength": 1
     },
+    "starts-at": {
+      "type": "string",
+      "description": "Starts at state name"
+    },
     "exec-status": {
       "type" : "string",
-      "enum": ["SYS.Timeout", "SYS.Fail", "SYS.Success", "SYS.InvalidParameter"],
-      "description": "Execution status"
+      "enum": ["Success", "Fail", "Timeout", "Invalid"],
+      "description": "Workflow execution status"
     },
     "trigger-defs": {
       "type": "array",
@@ -31,10 +35,6 @@
           {
             "title": "Delay State",
             "$ref": "#/definitions/delaystate"
-          },
-          {
-            "title": "End State",
-            "$ref": "#/definitions/endstate"
           },
           {
             "title": "Event State",
@@ -58,6 +58,7 @@
   },
   "required": [
     "name",
+    "starts-at",
     "states"
   ],
   "definitions": {
@@ -112,7 +113,7 @@
         },
         "timeout": {
           "type": "string",
-          "description": "Specifies the time period waiting for the events in the EVENTS-EXPRESSION"
+          "description": "Specifies the time period waiting for the events in the EVENTS-EXPRESSION (ISO 8601 format)"
         },
         "action-mode": {
           "type": "string",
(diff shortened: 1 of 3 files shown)


ruromero:
Why not making next-state optional? If there is no next, then it's an end :smile:

tsurdilo:
@ruromero i thought about that but in impls next-state can by dynamic (e.g process var). also its a string so not really reliable. explicit boolean end imo better.

cathyhongzhang:
If we remove the end state, then in the future if we need to extend the end state to hold more info/fields, we have to add those fields in each state that could possibly end the workflow rather than adding them in a centralized place. An explicit end state is better for extension and for debugging, I think. As to starts-at or start flag, I am not convinced that one is better than the other. Unless there is an obvious advantage, I do not see a point of changing the syntax.

tsurdilo:
@cathyhongzhang can you please explain:
"An explicit end state is better for extension and for debugging, I think" - what extension and debugging? Those are implementation details which implementors can decide, and most likely they will use existing state machine impls anyways which include these features.

IMO if we have an explicit end state then we should also have an explicit start state (your first point would count here as well), or not have both. Having no start state but an explicit end state makes no sense to me honestly.

ruromero:
I agree with both of you :smile: 

@tsurdilo I agree that if END state is explicit, the START state should also exist.
@cathyhongzhang it makes sense to me that having such states provide more flexibility in terms of extension. I don't worry about debugging for the reasons @tsurdilo mentioned.

I suggest we keep END and add START state. The start start can be triggered by events or manually so the only START states could be an EVENT state or a plain START state as the END one. That allow us to remove the `start` attribute from the other states.

tsurdilo:
Having an explicit start state is IMO confusing if it does not have to be used as a start of workflow. 
Don't think we have to worry about adding parameters to each state, as we already have exec-status defined as a workflow parameter. So in end I still think that removing end state is the better option of the two :)

ruromero:
In addition, depending on the workflow implemented which can be just a state machine there might not be an explicit end state.
I honestly don't have any clear position about this, yet.
@cathyhongzhang has a point saying we might want to add more fields there. But today, I can't think of any justification for having this END state. And if so, we should also add a 0..1 START state to allow a start of the workflow execution without an event interaction.

I think allowing the state transitions to carry out not only the message but also an execution state is more flexible. This execution state could be queried at any point and be modified given the intermediate states and not only "decided" at the END (SUCCESS/FAILURE).

In case a transformation is needed before ending the workflow it could be added as an OPERATION state or as an output filter in the last state.

Does that make sense?

tsurdilo:
@ruromero  what you are describing with optional start/end states that may do transformation sounds like a  script task in bpmn2, or a "Pass" state in step functions. I think there is a possibly need to add that but as its own state and not start/end. 

My opinion is that we should remove end state now and if it is needed or requested by community in the future with valid use case we can add it back but as the spec is right now it is clearly not needed.

ruromero:
Agree on removing end state. It is not justified ATM. As I mentioned in my previous comment, anything related to additional data or transformations can be done with existing states.

tsurdilo:
@ruromero thanks! sorry yeah i misunderstood your previous comment.

duglin:
all 3 of you should have merge rights now - so whoever is first can try to merge this  :-)
Just PLEASE only merge PRs for the workflow dir - nothing from the generic serverless wg stuff

cathyhongzhang:
Very good discussion! I do not have a strong point on either way.

tsurdilo:
rebased

PR #83, opened 2019-10-22
Update to switch state choices

tsurdilo:
Switch state choices had some errors, this fixes them.

* Single state should not be an array type
* Not state should not be an array type
* Updated required params
* Fixed typos
* Simplified operator enum values (there is no reason to have the "string-based" comparators...that seems just to be a copy of ones that are in aws step functions...). Also the english spelling makes more sense to readers.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -631,40 +631,27 @@
         },
         "operator": {
           "type": "string",
-          "enum": [
-            "EQ",
-            "LT",
-            "LTEQ",
-            "GT",
-            "GTEQ",
-            "StrEQ",
-            "StrLT",
-            "StrLTEQ",
-            "StrGT",
-            "StrGTEQ"
-          ],
+          "enum": ["Exists", "Equals", "LessThan", "LessThanEquals", "GreaterThan", "GreaterThanEquals"],
           "description": "Specifies how the input data is compared with the values"
         }
-      }
+      },
+      "required": ["path", "value", "operator"]
     },
     "singlechoice": {
       "type": "object",
       "description": "Single Choice",
-      "properties": {
-        "single": {
-          "type": "array",
-          "description": "List of choices",
-          "items": {
-            "type": "object",
-            "$ref": "#/definitions/defaultchoice"
+      "allOf": [
+        { "$ref": "#/definitions/defaultchoice" },
+        {
+          "properties": {
+            "nextState": {
+              "type": "string",
+              "description": "Specifies the name of the next state to transition to if there is a value match"
+            }
           }
-        },
-        "nextState": {
-          "type": "string",
-          "description": "State to transition to if there is a value match"
         }
-      },
-      "required": ["single", "nextState"]
+      ],
+      "required": ["nextState"]
     },
     "andchoice": {
       "type": "object",
@@ -711,7 +698,7 @@
         "not": {
           "type": "object",
           "$ref": "#/definitions/defaultchoice",
-          "description": "Single Choice"
+          "description": "Not Choice"
         },
         "nextState": {
           "type": "string",
--- a/workflow/spec/spec.md
+++ b/workflow/spec/spec.md
@@ -32,7 +32,6 @@ This document is a working draft.
 - [Examples](#Examples)
 - [Reference](#Reference)
 
-
 ## Introduction
 
 Serverless Workflow can be used to:
@@ -660,11 +659,10 @@ actions execute, a transition to "next state" happens.
 
 Switch states can be viewed as gateways. They define matching choices which then define which state should be 
 triggered next upon successful match.
+The current definitions of switch state defines multiple choices. And, Not, and Or choices
+define a single boolean operator which is to be applied to its elements. 
 
 #### <a name="switch-state-choices"></a>Switch State: Choices
-
-Switch states can be viewed as gateways. They define matching choices which then define which state should be 
-triggered next upon successful match.
 
 There are found types of choices defined:
 
@@ -678,7 +676,9 @@ There are found types of choices defined:
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| single |List of choices | array | yes |
+| path |Path that selects the data input value to be matched | string | yes |
+| value |Matching value | string | yes |
+| operator |Data Input comparator | string | yes |
 | [nextState](#Transitions) |State to transition to if there is valid match(es) | string | yes |
 
 <details><summary><strong>Click to view JSON Schema</strong></summary>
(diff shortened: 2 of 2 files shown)


workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -499,39 +499,26 @@
...
-            "LTEQ",
-            "GT",
-            "GTEQ",
-            "StrEQ",
-            "StrLT",
-            "StrLTEQ",
-            "StrGT",
-            "StrGTEQ"
-          ],
+          "enum": ["Equals", "LessThan", "LessThanEquals", "GreaterThan", "GreaterThanEquals"],
```
ruromero: What about adding NotEquals, Present and Blank/Empty ?
I can think of cases where someone would want to take decisions where a value is not present or empty. Besides, by adding NotEquals in many cases would simplify the workflow definitions.

NotChoice(SingleChoice:Equals(a,b)) --> SingleChoice:NotEquals(a,b)

workflow/spec/spec.md
```diff
@@ -637,6 +633,9 @@ There are found types of choices defined:
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | and |List of choices | array | yes |
+| path |Path that selects the data input value to be matched | string | yes |
```
ruromero: The changes do not affect the and choice.
```
{
  "and": [default-choices]
  "next-state": "state-a"
}

File: workflow/spec/schema/serverless-workflow-schema-01.json [hunk 3]
```diff
@@ -499,39 +499,26 @@
...
+            }
           }
-        },
-        "next-state": {
-          "type": "string",
-          "description": "Specifies the name of the next state to transition to if there is a value match"
         }
-      }
+      ],
+      "required": ["path", "value", "operator", "next-state"]
```
ruromero: shouldn't be "path", "value", "operator" be set as required in defaultchoice instead and here only require next-state?

workflow/spec/spec.md
```diff
@@ -680,20 +679,22 @@ There are found types of choices defined:
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| not |List of choices | array | yes |
+| not |Choice | object | yes |
+| path |Path that selects the data input value to be matched | string | yes |
```
ruromero: path, value, operator are part of the defaultchoice.

workflow/spec/spec.md
```diff
@@ -714,7 +715,7 @@ There are found types of choices defined:
             "description": "Specifies the name of the next state to transition to if there is a value match"
         }
     },
-    "required": ["not", "next-state"]
+    "required": ["not", "path", "value", "operator", "next-state"]
```
ruromero: Remove

workflow/spec/spec.md
```diff
@@ -725,14 +726,17 @@ There are found types of choices defined:
 | Parameter | Description |  Type | Required |
 | --- | --- | --- | --- |
 | or |List of choices | array | yes | 
+| path |Path that selects the data input value to be matched | string | yes |
```
ruromero: Remove path, value,operator as they are part of the default choice (inside or)

workflow/spec/spec.md
```diff
@@ -758,7 +762,7 @@ There are found types of choices defined:
             "description": "Specifies the name of the next state to transition to if there is a value match"
         }
     },
-    "required": ["or", "next-state"]
+    "required": ["or",  "path", "value", "operator", "next-state"]
```
ruromero: Revert

workflow/spec/spec.md
```diff
@@ -670,7 +669,7 @@ There are found types of choices defined:
             "description": "Specifies the name of the next state to transition to if there is a value match"
         }
     },
-    "required": ["and", "next-state"]
+    "required": ["and", "path", "value", "operator", "next-state"]
```
ruromero: Remove

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: For "Not", "And", "Or" you use the appropriate choices defined. So for NotEquals you have the "Not Choice" with "Equals" operator.

workflow/spec/spec.md
tsurdilo: defaultchoice is no longer a valid choice (only used in json schema for extending params purposes). There is and not or and single choices. Sorry maybe I don't understand the change requested here.

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: updated. good point

workflow/spec/spec.md
tsurdilo: yes but these tables show all properties that users can define. defaultchoice in the schema defaultchoice is just used for extending params so we dont have to duplicate them across all valid choices.

workflow/spec/spec.md
tsurdilo: there is a difference between the json schemas for thisshown in the spec.md and serverless-workflow-schema-01.json and this is done on purpose. Since defaultchoice is not an actual choice in spec.md we list all of its properties in each choice so readers can see them (same for required params). The actual serverless-workflow-schema-01.json which implementors would use has the 

"$ref": "#/definitions/defaultchoice",

this makes this section of spec.md readable. Readers would be otherwise confused and not see all properties that are available.

workflow/spec/spec.md
tsurdilo: see comment above

workflow/spec/spec.md
tsurdilo: same comment about defaultchoice not being an actual choice as above.

workflow/spec/spec.md
ruromero: Given the current schema. You cannot define `path` next to `and`. It has to go inside the choices.
```
{
  "and": [
    {
      "path": "$.name",
      "condition": "Equals",
      "value": "Bob"
    }, {
     "path": "$.age",
     "condition": "GreaterThan",
     "value": "18"
    }
  ],
  "next-state": "the-next-state"
}

File: workflow/spec/spec.md (hunk 8 above)
Comment (tsurdilo) [reply workflow/spec/spec.md]: Why. spec.md shows the "resolved" json schema. it does not include $ref. This is done so readers can see all of the fields. There is a link in spec.md to the full schema with all the references where refs are used to defaultchoice for param reuse.

File: workflow/spec/spec.md (hunk 2 above)
Comment (tsurdilo) [reply workflow/spec/spec.md]: Why should that be an issue? Each "and" element can have own path. Same for "or", both being array types.

**Comment by ruromero** [comment]
@tsurdilo I think the whole approach is incorrect and a simple set of examples will help us understand it. A switch is a set of choices and each choice can take you to a different path. But the problem is that now each choice is a single AND/OR/NOT/Single.
Example:

```
default: "default-state"
choices:
 - and: 
     - value: 'Bob'
       condition: Equals
       path: $.name
     - value: 18
       condition: LessThanEquals
       path: $.age
   next-state: 'rejected'
 - value: 'VIP'
   condition: Equals
   path: $.ticket
   next-state: 'granted'
```

See? There is no way to create complex conditions like NOT(AND(cond1, cond2)) / OR(AND(cond1, cond2), NOT(cond3), AND(cond4, cond5))

I have created a PR to your branch with my suggestion. Let me know what do you think. https://github.com/tsurdilo/wg-serverless/pull/1

**Comment by tsurdilo** [comment]
@ruromero yes I agree with you. We should revise the entire state. This pr was supposed to be just the fix for the way it currently is. I added to this pr for documentation purposed:

```
The current definitions of switch state defines multiple "choices". And, Not, and Or choices
define a single boolean operator which is to be applied to its elements. 
We will lift this limitation to allow a more flexible approach in the near future.
```

I suggest add this as it fixes whats there in master now. Then we create a discussion thread under "Issues" (since there is no other place really atm) and start figuring out what to do with this state (based on your idea in the branch you created) and create a new pr with a bigger update. WDYT?

File: workflow/spec/schema/serverless-workflow-schema-01.json [hunk 9]
```diff
@@ -499,39 +499,27 @@
...
       "description": "Single Choice",
-      "properties": {
-        "single": {
-          "type": "array",
-          "description": "List of choices",
-          "items": {
-            "type": "object",
-            "$ref": "#/definitions/defaultchoice"
+      "allOf": [
+        { "$ref": "#/definitions/defaultchoice" },
```
cathyhongzhang: I was confused by "allof ... defaultchoice". When I took a look at the defaultchoice, I understand it. Maybe we should not call it "defaultchoice" since it is not a choice. How about giving it another name, such as "conditioncheck" or something like this, to avoid confusion.

ruromero: @cathyhongzhang I have an alternative proposed as mentioned in my comments. I am waiting for @tsurdilo to merge it to his branch.

tsurdilo:
rebased against master

tsurdilo:
Hi, this pr includes fixes to the current master spec doc. It has been idle for over a month and the updates to the pr branch have been removed over 20 days ago. 
Since this fixes currently existing errors in specification I will push this. Further modification proposals can be done via new prs. Thanks.

PR #84, opened 2019-10-24
Update retry in action definition

tsurdilo:
This updates the retry definition of actions. Main reason for update is that currently retry is a single element and it should be an array of retry definitions each matching on the return of the function call.

* Renamed to retry-when and is an array of retry definitions
* Retry definition parameters: match, interval, max-retry, next-state
* Added "default" parameter to Action definition. This defines the state to transition to if the function call fails (error) and none of the retry definitions define a match for it.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -193,10 +193,6 @@
         },
         "filter": {
           "$ref": "#/definitions/filter"
-        },
-        "default": {
-          "type": "string",
-          "description": "State to transition to in case there are errors and no retry definitions have been triggered"
         }
       },
       "required": ["function"]
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -183,7 +183,7 @@
           "type": "string",
           "description": "Specifies the time period waiting for the events in the eventExpression (ISO 8601 format)"
         },
-        "retry-when": {
+        "retry": {
           "type": "array",
           "description": "Retry Definitions",
           "items": {
@@ -196,7 +196,7 @@
         },
         "default": {
           "type": "string",
-          "description": "State to transition to in case there are errors and no retry-when definitions have been triggered"
+          "description": "State to transition to in case there are errors and no retry definitions have been triggered"
         }
       },
       "required": ["function"]
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -211,13 +211,13 @@
         },
         "max": {
           "type": "integer",
-          "default":"0",
+          "default": "0",
           "minimum": 0,
           "description": "Specifies the max retry"
         },
         "nextState": {
           "type": "string",
-          "description": "State to transition to when exceeding maxRetry limit"
+          "description": "State to transition to when exceeding max limit"
         }
       },
       "required": ["match", "nextState"]
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -183,13 +183,20 @@
           "type": "string",
           "description": "Specifies the time period waiting for the events in the eventExpression (ISO 8601 format)"
         },
-        "retry": {
-          "type": "object",
-          "$ref": "#/definitions/retry",
-          "description": "Specifies how the result from a function is to be handled"
+        "retry-when": {
+          "type": "array",
+          "description": "Retry Definitions",
+          "items": {
+            "type": "object",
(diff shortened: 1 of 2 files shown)


tsurdilo:
ill fix merge conflicts once a decision on this has been made (before merging)

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -84,25 +84,6 @@
       },
       "required": ["name", "source", "type"]
     },
-    "filter": {
```
ruromero: Why did you remove filters?

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -154,13 +132,17 @@
           "minimum": 0,
           "description": "Specifies the time period waiting for the events in the EVENTS-EXPRESSION"
         },
-        "retry": {
-          "type": "object",
-          "$ref": "#/definitions/retry",
-          "description": "Specifies how the result from a function is to be handled"
+        "retry-when": {
```
ruromero: A `default` parameter inside an `action` is not easy to be understood as a the default state to go when an error happened.
I suggest you create the following:

* Retry has an array of `match`es and a `default` which will be easily understood as a default match.
* A `match` will then have a `interval`, `max-retry`, `next-state` and instead a `match` I suggest adding a `condition` (see https://github.com/tsurdilo/wg-serverless/pull/1). This will align with the existing way of defining conditions. Currently "match" uses a language expression. We can also use this for event-expressions. WDYT?

```
retry:
  when:
   - match:
     ...
   - match:
     ...
  default: "default-error-state"
```

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: this pr was done against master. we had the filters defined in the json schema but not in the spec document. will fix this. sry

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: For your example "when" would need to be an array so this can't be described in schema as shown 
Would you be ok with:
```
"retry": {
  "when": [
     {
       ...
     },
     {
       ...
     }
  ]
}
```

ruromero: Yes, that was my point. I oversimplified the example and used yaml. Sorry :smile: 
Do you agree with this?

workflow/spec/spec.md
```diff
@@ -340,8 +340,9 @@ Each event state's event definition includes one or more actions. Let's define t
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | [function](#Function-Definition) |Function to be invoked | object | yes |
-| timeout |Max amount of time in seconds to wait for the completion of the function's execution | integer | no |
-| [retry](#Retry-Definition) |Defines if funtion execution needs a retry | object | no |
+| timeout |Max amount of time to wait (in seconds) for the completion of the function's execution | integer | no |
+| [retry-when](#Retry-Definition) |Defines function retries | array | no |
```
cathyhongzhang: The name "retry-when" is a little misleading since the semantics is not just about "when". It seems better to just call it"retry" to match the stated description

tsurdilo: Ruben suggested a solution in above comment that would make this:
```
"retry": {
   "when": {
     "match": "abc",
...
 },
{
"match": "xyz",
...
}
}
```
the "when" then is assocated with matches so like in language form would be "Retry when match is "abc" and "Retry when match is xyz". 

Would you be ok with that?

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: yes

workflow/spec/spec.md
cathyhongzhang: I see this is closed? Are you going to proceed with this change?

tsurdilo: @cathyhongzhang pr is open -- can you let us know if rubens suggestion in comment above is ok with you ?

cathyhongzhang: Thanks for checking with me! I think it is better to just use "retry" since 1) the semantics is not just about "when", 2) the syntax is simpler than a two-level syntax. What do you think?

tsurdilo:
rebased and fixed typos

workflow/spec/spec.md
tsurdilo: sure i have changed to just "retry" without the "when". agree its simpler

tsurdilo:
I have updated to changes requested by both ruben and cathy. This pr has been sitting idle without comments for over a month so I will assume noone has anything else to contribute to it. 

This fixes current issues in master where retry is not array type. 

Will push this pr through as it fixes things. Further changes can be proposed via future prs.

PR #85, opened 2019-10-28
Changing time values from seconds to ISO 8601 format

tsurdilo:
time based properties such as "timeout" and "retry-interval" are defined as "in seconds" which is not very user friendly. instead we should use the ISO 8601 dat and time format -- https://www.iso.org/iso-8601-date-and-time-format.html

which is international standard and much more intuitive to represent time and delays etc.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -112,7 +112,7 @@
         },
         "timeout": {
           "type": "string",
-          "description": "Specifies the time period waiting for the events in the EVENTS-EXPRESSION"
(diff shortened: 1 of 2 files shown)


tsurdilo:
I'm planning to update the spec document to add in addition to 
"Click to view JSON Schema"
also 
"Click to view Example JSON" where I will add examples for these there. WDYT?

ruromero:
I think it will be very useful to readers

tsurdilo:
@ruromero ok will add. just trying to see where as we dont have examples for any of the parameters in those tables currently. maybe add a new column

tsurdilo:
@ruromero updated to add examples. They are in the Description column on the property tables.

cathyhongzhang:
LGTM

tsurdilo:
fixed typos.

PR #86, opened 2019-10-28
Adding filter definition

tsurdilo:
The document was missing filter definition which is described in the "Filter Mechanism" section. This adds it.

### DIFF
--- a/workflow/spec/spec.md
+++ b/workflow/spec/spec.md
@@ -153,7 +153,7 @@ Here we define details of the Serverless Workflow definitions:
             "description": "Trigger Definitions",
             "items": {
                 "type": "object",
-                "$ref": "#definitions/triggerevent"
+                "$ref": "#/definitions/triggerevent"
             }
         },
         "states": {
@@ -250,6 +250,7 @@ We will start defining each individual state:
 | type |start type | string | yes |
 | end |Is this state an end state | boolean | no |
 | [events](#eventstate-eventdef) |Array of event | array | yes |
+| [filter](#Filter-Definition) |State data filter | object | yes |
  
 <details><summary><strong>Click to view JSON Schema</strong></summary>
 <p>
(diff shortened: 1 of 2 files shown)


tsurdilo:
ill fix merge conflicts once a decision on this has been made (before merging)

tsurdilo:
@ruromero rebased

ruromero:
@cathyhongzhang any comments?

cathyhongzhang:
@ruromero Thanks for checking with me. I am in meetings and will find time to review it this afternoon.

workflow/spec/spec.md
```diff
@@ -482,15 +496,18 @@ as well as define parameters (key/value pairs).
             "description": "Actions Definitions",
             "items": {
                 "type": "object",
-                "$ref": "#definitions/action"
+                "$ref": "#/definitions/action"
             }
         },
+        "filter": {
+          "$ref": "#/definitions/filter"
```
cathyhongzhang: If I read the syntax correctly, it does not seem to allow applying different filters to different actions. The original spec allows that.

workflow/spec/spec.md
```diff
@@ -911,6 +940,42 @@ The elements of the output array need not be of the same type.
...
+      "type": "string",
+      "description": "Specify result JSON node of Action Output as JSONPath"
+    },
+    "output-path": {
+      "type": "string",
+      "description": "Specify output data of State or Action as JSONPath"
+    }
+  },
+  "required": ["input-path", "result-path", "output-path"]
+}
```
cathyhongzhang: All three are required, but for an event, we only need input-path filter. Could you clarify? Thanks

workflow/spec/spec.md
tsurdilo: @cathyhongzhang each action definition includes a filter property. so the functionality you mention is preserved. The operation state itself also needs a filter definitions which is what you are pointing at.

workflow/spec/spec.md
tsurdilo: good point. fixed

tsurdilo:
i will rebase this in case the camel case props pr gets pushed before this

workflow/spec/spec.md
cathyhongzhang: I see. Thanks for the clarification. I thought this filter is for the action.

workflow/spec/spec.md
cathyhongzhang: LGTM

ruromero:
I will merge it then once #96 is merged

PR #90, opened 2019-10-31
Adding extension capability to workflow model

tsurdilo:


workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -54,6 +54,13 @@
           }
         ]
       }
+    },
+    "extensions": {
+      "type": "array",
+      "description": "Workflow Extensions",
+      "items": {
+        "type": "object"
+      }
```
cathyhongzhang: I don't feel that we have enough use cases to know how to define and justify the definition of the extension. In order for this spec to be widely adopted, it is better to define the basics that a workflow must have and leave other parts open for others to contribute . If we define too much, it may make people feel like a complete and closed solution. Hope the spec can have a bigger community to contribute, which is the key for a project to succeed in CNCF.

cathyhongzhang: NA

tsurdilo: @cathyhongzhang I don't think maybe you are getting what this change is about. 

I'll try to clear up the confusion about comment -- "In order for this spec to be widely adopted, it is better to define the basics that a workflow must have and leave other parts open for others to contribute."

This PR does provide the ability for community to define and add their contributions. Without this, users would have to go outside the workflow specification to build extensions. With this they can:
1. Build serverless workflow specific extensions
2. Share workflow extensions with community 
3. Not have to define separate json for workflow and their workflow-specific extensions

BPMN2 also had an extension mechanism in place right from the start. And there are bpmn2 extensions such as BPSim that are widely used. 

Regarding comment --"If we define too much, it may make people feel like a complete and closed solution" 
How can this change make something "closed" and "complete" when it gives users ability to define what they want how they want it? To me it's the other way around: Unless we provide extension capabilities, users are forced to only use what we give them which THEN makes the specification "closed" and "complete".

tsurdilo: I seriously think we should add to the review rules:

"You cannot vote NA unless you really understand what the change is about". Otherwise really even contributing to this is completely useless......

ruromero: I will add my two cents here.
The way I see it, this indeed doesn't close the specification. But at this early stage I don't consider this functionality justified. The fact that BPMN2 has it is not enough.

I believe that if that opening the door to extensions will not help this spec to be properly defined as each implementation will workaround the gaps through extensions. If we allow contributors and users to understand the existing possibilities and find a common use case that requires the spec to change then it will naturally appear and be discussed.

However, if with the time we see that there are common use cases that justifies some sort of extensions, then we can review this proposal again.

Not having extensions doesn't mean the spec is closed. If there's a use case the spec can't cover we MUST understand why, discuss it and agree or not but not just allow anyone to do it in its own way.

tsurdilo: regarding @ruromero comments ...what? We have used bpmn2 spec (which  has been around over a decade and is 10x and more bigger than this, and has huge adoption) for a long time and still look how bpmn2 looks for example:

```xml
<bpmn2:scriptTask id="_1FBF9BD4-C6E4-4411-8CC9-5944CD34183B" drools:selectable="true" name="Hello world (1)" scriptFormat="http://www.java.com/java">
    ...

<bpmn2:parallelGateway id="_08822C48-0876-44C5-AC17-04504FAE9B78" drools:bgcolor="#f0e68c" drools:selectable="true" drools:bordercolor="#a67f00" name="hello_gateway_out" gatewayDirection="Converging">
    ....

<bpmn2:relationship id="_8Wkj_pKGEeSzKuqwkg58Fg" type="BPSimData">
    <bpmn2:extensionElements>
      <bpsim:BPSimData>
       ....     
```

Same is for BPMN2 produced for camunda. All the "drools" and "bpsim" are namespaces which are used so we can extend the BPMN2 specification parameters. The relationship element allows for things like extensions. 

I am showing you this because even for a specification like BPMN2 to make it executable you need to extend it..just think how it will be here with our little 5 states and 10 parameters :) 

With JSON we cannot do parameter extension (but ok we can add metadata array), but having extension capability should not be seen as a "workaround" imo.

tsurdilo: And before you two say something like well I am just trying to copy bpmn2 spec, I'll say it right away that is not true. What I do know is how to go from a specification (much much larger than this one) to an executable engine using it. Also I know how to go from a specification to usable tooling support for it (again to make it executable). Worked on that for many years. 
So in some ways I see based on experience what we are missing and things that we highly use in other similar specs (such as bpmn2..yes) and try to integrate their ideas here so we don't run into issues later. Thats all.

tsurdilo:
@cathyhongzhang ok I understand the point of NA. Thats fine. 
What I think we are still not on the same page is that you think we are "defining a specific extension". That's not the case, what we are defining is the **ABILITY** for users to define their own extension points. There is not much we have to IMO worry here as:

1. Users can just ignore and not have the need to define an extension
2. Users have complete freedom to define anything they want in the "extensions" element.

So we are not forcing anything to anyone, only thing this does is give more choice to users up-front instead of seeing this specification as "closed" and "completed" without an ability to add their own ideas and mind into it.

tsurdilo:
@cathyhongzhang And about your statement on the "community needs" above -- we have a comment here from @salaboy who **IS** a community member and has stated that this would be useful to him. And in addition he is by far the most knowledgeable of any of us here regarding workflows and their implementations :) So you kinda have an opinion of the community but let's not use it ... :)

salaboy:
@tsurdilo I think that we should define the scope of the spec.. as I believe that extensions will be a core building block that will need to be defined sooner rather than later.

mswiderski:
I also agree with @tsurdilo and @salaboy on the importance of extension. I'd like to abstract us a bit from BPMN2 because I believe this is starting to compare two workflow formats while it is not really relevant in this context. Instead have a look at cloud events specification. They defined (not detailed) extension mechanism because in the end this is what is a must for pretty much any implementation of any specification. Stating that it's too early is like saying let's way for someone to complain they can't provide extended/enhanced implementation for the spec and then try to figure out what that person/project needs which would usually lead to opinionated definition of the extension. So +1 for this PR to allow to add extra data/information on top of well defined constructs.

ruromero:
I understand your points and I agree. In addition thanks to how json schema validation works users will be able to provide additional validations to the extensions content while still being a valid workflow.
Thank you for your comments.

LGTM

cathyhongzhang:
Got it. So here we are just defining a "property bag" called "extension" and we are not defining any specific extension property/item yet (those are just examples). I just want to make sure we are on the same page that if someone likes to define a specific extension, it should have a clear and "common enough" use case to justify it being put into the spec. I remember we had several weeks of discussion  and different opinions on how to position and define extension in CloudEvents. I would like this PR to add some description similar to "Extension Context Attributes" of the CloudEvents spec.md and "CloudEvent Attribute Extensions" of the primer.md.

tsurdilo:
@cathyhongzhang yes that is correct. Added "Extending" section to the main spec documentation which contains a better explanation and link to the spec-extending markup doc which includes more info and an example. Hope this helps.

tsurdilo:
rebased

tsurdilo:
@cathyhongzhang could you decide if we can lift your "NA" or do you need more info/time to decide if we can include this or not.

cathyhongzhang:
Sorry forgot to lift this. Done

PR #92, opened 2019-10-31
Adding Reference Section 

salaboy:
This PR adds the reference section with a mention to the BPMN standard by the OMG group as well as encourage people to share other related tools that can help to make decisions while creating this specification. 

I've included some examples covering the use cases listed in this spec translated into BPMN for further comparison.


ruromero:
I don't think the bpmn workflows are need. I consider the images of the generated processes should suffice.

salaboy:
@ruromero I think that there is value in having files to compare both, graphical and technical description of the models. I am happy to remove, but the files also give you pointers to the schemas which makes the definitions easy to validate.

tsurdilo:
I think having the bpmn xml in separate files and can be useful to readers. 

@salaboy the only thing I'd suggest is to remove from the bpmn2 xml:
exporter="Camunda Modeler" 
exporterVersion="3.3.5"

so we stay vendor neutral ;)

salaboy:
@tsurdilo apologies.. done!

tsurdilo:
LGTM

PR #94, opened 2019-11-01
Updating state transitions

tsurdilo:
This PR updates the workflow state transitions.
Currently state transitions are based on the "next-state" property which should match an existing state name. Issues with that are:
* State name should be primarily used for display purposes (in notation for example)
* State name can include "special" (i18n) characters which can cause issues for implementations.
* There could be multiple states with the same name if the users chose to

So overall "name" property of a state does not guarantee uniqueness and can lead to transition problems/ambiguity.

In addition, the current transition based on "next" or "next-state" is exactly the same as in aws step functions. Even if that is probably a good strategy there, step functions have closed system implementation where they can guarantee uniqueness of the state name. For a specification which targets multiple implementations, that is not a good thing to enforce.

This pr:
* adds a not-required "id" property to each state.
* adds a "Transitions" section which describes the options implementations have in order to assure
the uniqueness and resolve ambiguities of state transitions.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -158,16 +158,16 @@
             "$ref": "#/definitions/action"
           }
         },
-        "nextState": {
+        "target": {
           "type": "string",
-          "description": "Name of the next state to transition to after all the actions for the matching event have been successfully executed",
+          "description": "State to transition to after all the actions for the matching event have been successfully executed",
           "minLength": 1
         },
         "filter": {
           "$ref": "#/definitions/filter"
         }
       },
-      "required": ["eventExpression", "timeout", "nextState"]
+      "required": ["eventExpression", "timeout", "target"]
     },
     "action": {
       "type": "object",
@@ -208,9 +208,9 @@
           "minimum": 0,
           "description": "Specifies the max retry"
         },
-        "nextState": {
+        "target": {
           "type": "string",
-          "description": "Name of the next state to transition to when exceeding maxRetry limit",
+          "description": "State to transition to when exceeding max-retry limit",
           "minLength": 1
         }
       }
@@ -281,9 +281,13 @@
       "type": "object",
       "description": "Causes the workflow execution to delay for a specified duration",
       "properties": {
+        "id": {
+          "type": "string",
+          "description": "State id"
+        },
         "name": {
           "type": "string",
-          "description": "Unique name of the state"
+          "description": "State name"
         },
         "type": {
           "type": "string",
@@ -304,9 +308,9 @@
           "type": "string",
           "description": "Amount of time (ISO 8601 format) to delay"
         },
-        "nextState": {
+        "target": {
           "type": "string",
-          "description": "Name of the next state to transition to after all the delay."
+          "description": "State to transition to after all the delay."
         }
       }
     },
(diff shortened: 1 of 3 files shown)


tsurdilo:
I will resolve conflicts once a decision on this has been made.

cathyhongzhang:
I will review this tomorrow

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -130,16 +130,16 @@
             "$ref": "#/definitions/action"
           }
         },
-        "next-state": {
+        "target": {
           "type": "string",
```
cathyhongzhang: I am not convinced of this change. I do not see anything wrong with "next-state", which is self explanatory.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -253,9 +253,15 @@
       "type": "object",
       "description": "Causes the workflow execution to delay for a specified duration",
       "properties": {
+        "id": {
+          "type": "string",
+          "description": "State id",
+          "pattern": "$[a-zA-Z0-9\\-\\.]+^",
+          "minLength": 1
+        },
```
cathyhongzhang: Could you clarify the reason to allow some special characters, but not the other? Do we need to specifically define the ID pattern? In CloudEvents, there is no such restriction.

tsurdilo: I checked in cloudevents they use UUID. This was what I had in first iteration then it was changed to a pattern. I agree with using UUID btw :) so would be happy to change it back if thats what you guys want.

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: I'm ok with calling it next-state or nextState, but the issue still remains that it is a string and as can pose issues for implementations to depends on uniqueness solely on it. That's why this PR adds an id element (uuid) which is much stronger for lookups.

tsurdilo: if we want to keep next-state lookup only thats fine w/ me as well and can keep this PR around for later evaluation. 

Just to give another example, it is difficult to do string comparisons (in many languages based on string values, as often many different character sets are involved etc). Wherever we have a definition of "unique value" within a workflow we should consider using UUID (same is for cloudevents)

cathyhongzhang:
I suggest that we remove those changes  regarding next-state to target. I am OK with adding an optional ID field ( "UUID") to the state to avoid string comparison in implementation.

tsurdilo:
@cathyhongzhang changed back "target" to "nextState", and removed the pattern definition (so implementors can use uuid). I'll rebase if approved.

workflow/spec/spec.md
```diff
@@ -441,12 +449,12 @@ as well as define parameters (key/value pairs).
...
         },
-        "next-state": {
+        "nextState": {
             "type": "string",
-            "description": "Name of the next state to transition to when exceeding max-retry limit"
+            "description": "State to transition to when exceeding max-retry limit"
         }
     },
-    "required": ["match", "next-state"]
+    "required": ["match", "nexttState"]
```
ruromero:
```suggestion
    "required": ["match", "nextState"]
```

cathyhongzhang: LGTM after you fix the typo Ruben pointed out.

tsurdilo: fixed. will rebase and merge. thanks guys.

PR #95, opened 2019-11-02
Adding workflow properties

tsurdilo:


tsurdilo:
Adding properties (not required) to workflow:
* title - title of the workflow. can be used by notation and tooling
* description - workflow description - describes the purpose of the workflow, also can be used by notation and tooling
* version - title of workflow, helps with content management and execution 
* last-updated - last updated date+time of workflow json. helps with content management and execution
* created - workflow json creation date + time. helps with content management and execution

also adding "additionalProperties": false which is a json schema specific property to not allow properties that are not listed as defined in the workflow schema. This is for validation.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -9,6 +9,31 @@
...
+      "description": "Workflow description"
+    },
+    "version": {
+      "type": "string",
+      "description": "Workflow version",
+      "minLength": 1
+    },
+    "last-updated": {
+      "type": "string",
+      "description": "Workflow last updated date and time"
```
cathyhongzhang: Since there is already a workflow version, why do we need last update and create date and time? Could you give an example of why the users care about these info?

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -9,6 +9,31 @@
       "description": "Workflow name",
       "minLength": 1
     },
+    "title": {
```
cathyhongzhang: What is the difference between workflow name and title?

workflow/spec/spec.md
```diff
@@ -157,7 +188,8 @@ Here we define details of the Serverless Workflow definitions:
             }
         }
     },
-    "required": ["name", "states"]
+    "required": ["name", "states"],
+    "additionalProperties": false
```
cathyhongzhang: Could you clarify what this addtionalProperties mean?

tsurdilo: i put the description in the pr initial comment above --
"additionalProperties": false which is a json schema specific property to not allow properties that are not listed as defined in the workflow schema. This is for validation.

It basically means "users cannot add their own made-up properties to the workflow json. only the ones defined in the schema are allowed.

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: Since we do not add an unique id for workflow I assume most people will make name its unique ascii string identifier. Title on the other hand can be used by tooling to display and i18n chars sets can be used.

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: it just gives more information about the workflow to users. if hey don't care they wont look at  it. if they care they can set it .

tsurdilo: I'm ok with removing these two if you guys feel they are not necessary.

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: I'm ok with removing "title" if you guys feel that "name" and "description" are enough.

ruromero:
I think `schema-version` and `version` are useful, specially the first one. On the contrary, I don't see the need of allowing any other properties. That's we initially defined the metadata, didn't we?
Control version systems already provide the possibility to see the creation/update of a file and as mentioned already, `title` is similar to `name`.

tsurdilo:
metadata is not in the specification so cant use it. you will need to be able to uniquely distinguish a workflow definition. since this spec does not include (and it should) a workflow unique id and also id properties for all states etc imo) you kinda fall back here on name+version which is weak.

ruromero:
> metadata is not in the specification so cant use it. you will need to be able to uniquely distinguish a workflow definition. since this spec does not include (and it should) a workflow unique id and also id properties for all states etc imo) you kinda fall back here on name+version which is weak.

The workflow definition has a metadata in the `workflow-api`. I thought it was here too, my bad.

Regarding ID/name/title/description fields. I suggest the following:
* We need a field to name a workflow in a computer-standardized way. For that we can have an `id` that allows only [a-zA-Z0-9\-\.]
* We need a human readable name. We keep `name`
* We need a human readable description of what the process does. We keep `description`

tsurdilo:
ok from recommendations i did updates:
* added "id" parameter - there is no way in json schema to describe UUID format so its a string but did put in in description that it has to conform to UUID
* removed created and last updated properties
* removed title property

tsurdilo:
rebased

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -61,6 +79,7 @@
     "starts-at",
     "states"
   ],
+  "additionalProperties": false,
```
ruromero: Remove this as this cannot be used in json schema validation. Besides, the proposed new properties are optional and self-descriptive

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -4,11 +4,29 @@
   "description": "Workflow is a vendor-neutral specification for defining the format/primitives that the users can use to specify/describe their serverless application flow.",
   "type": "object",
   "properties": {
+    "id": {
+      "type": "string",
+      "description": "Workflow unique identifier. Conforms the universally unique identifier (UUID) format.",
```
ruromero: We should not enforce UUID format. As long as it has simple characters 
```suggestion
      "pattern": "$[a-zA-Z0-9\-\.]+^",
      "description": "Workflow identifier. Accepted characters include both uppercase and lowercase ASCII letters, the digits 0 through 9, space, dash and dot: A-Z, a-z, 0-9, -, .",
```

workflow/spec/schema/serverless-workflow-schema-01.json
cathyhongzhang: I don't think these are necessary in the main spec

workflow/spec/spec.md
```diff
@@ -147,7 +170,8 @@ Here we define details of the Serverless Workflow definitions:
             }
         }
     },
-    "required": ["name", "starts-at", "states"]
+    "required": ["id", "name", "starts-at", "states"],
+    "additionalProperties": false
```
cathyhongzhang: I feel the workflow version should be required. My understanding of the version defined here is the version of a specific user's workflow spec (e.g. a user might have multiple versions of a home monitoring app spec), right?

tsurdilo:
@ruromero fixed. (had to escape the regex pattern so its valid json tho)

workflow/spec/spec.md
tsurdilo: ok. fixed.

workflow/spec/spec.md
```diff
@@ -147,7 +171,7 @@ Here we define details of the Serverless Workflow definitions:
             }
         }
     },
-    "required": ["name", "starts-at", "states"]
+    "required": ["id", "name", "version", "starts-at", "states"]
 }
 ```
```
cathyhongzhang: LGTM

PR #97, opened 2019-11-05
Proposal for new state: SubFlow State

tsurdilo:
State Description (also included in spec.md with this change):
********
It is often the case that you want to group your workflows into small, **reusable** logical units that perform certain needed functionality.
Even tho you can use the Event state to call an externally deployed services (via function), at times
there is a need to include/nest another serverless workflow (from classpath/local file system etc). In that case you would use the SubFlow State.
It also allows users to model their workflows with reusability and logical grouping in mind.

This state allows you to include/nest a uniquely identified sub-workflow and start its execution. 
Another use of this state is within [branches](#parallel-state-branch) of the [Parallel State](#Parallel-State). Instead of having to define all states
in each branch, you could separate the branch states into individual sub-workflows and call the SubFlow state
as a single state in each.

Sub-workflows must have a defined start and end states. 
The wait-for-completion property defines if the SubFlow state should wait until execution of the sub-workflow
is completed or not. 

Each sub-workflow receives a copy of the SubFlow state's input data.
If wait-for-completion property is set to true, sub-workflows have the ability to edit the parent's workflow data.
If this property is sete to false, data access to parent's workflow should not be allowed.
********

Similar states used in other "workflow" solutions:
BPMN2: "Call Activity"
Netflix Condutor: "Sub Workflow"

Depends on https://github.com/cncf/wg-serverless/pull/95 (where we add workflow "id" property) and https://github.com/cncf/wg-serverless/pull/86 (where we add the filter definition).

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -72,7 +72,7 @@
             "$ref": "#/definitions/switchstate"
           },
           {
-            "title": "Switch State",
+            "title": "Invoke State",
             "$ref": "#/definitions/invokestate"
           }
         ]
@@ -271,7 +271,7 @@
                 "$ref": "#/definitions/switchstate"
               },
               {
-                "title": "Switch State",
+                "title": "Invoke State",
                 "$ref": "#/definitions/invokestate"
               }
             ]
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -70,6 +70,10 @@
           {
             "title": "Switch State",
             "$ref": "#/definitions/switchstate"
+          },
+          {
+            "title": "Switch State",
+            "$ref": "#/definitions/invokestate"
           }
         ]
       }
@@ -265,6 +269,10 @@
               {
                 "title": "Switch State",
                 "$ref": "#/definitions/switchstate"
+              },
+              {
+                "title": "Switch State",
+                "$ref": "#/definitions/invokestate"
               }
             ]
           }
@@ -503,6 +511,47 @@
           "description": "Specifies the name of the next state if there is no match for any choices value"
         }
       }
+    },
+    "invokestate": {
+      "type": "object",
+      "description": "Defines a local workflow to be executed",
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "Unique name of the state"
+        },
+        "type": {
+          "type" : "string",
+          "enum": ["INVOKE"],
+          "description": "State type"
+        },
+        "end": {
+          "type": "boolean",
+          "default": false,
+          "description": "Is this state an end state"
+        },
+        "wait-for-completion": {
+          "type": "boolean",
+          "default": false,
+          "description": "Workflow execution must wait for local workflow to finish before continuing."
+        },
+        "workflow-id": {
+          "type": "string",
+          "description": "Local workflow unique id."
+        },
+        "workflow-version": {
+          "type": "string",
+          "description": "Local workflow version"
+        },
+        "filter": {
+          "$ref": "#/definitions/filter"
+        },
+        "next-state": {
+          "type": "string",
+          "description": "Specifies the name of the next state to transition to after local workflow has completed execution."
+        }
+      },
+      "required": ["name", "type", "next-state", "workflow-id"]
     },
     "defaultchoice": {
       "type": "object",
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -72,8 +72,8 @@
             "$ref": "#/definitions/switchstate"
           },
           {
-            "title": "Invoke State",
-            "$ref": "#/definitions/invokestate"
+            "title": "SubFlow State",
+            "$ref": "#/definitions/subflowstate"
           }
         ]
       }
@@ -271,8 +271,8 @@
                 "$ref": "#/definitions/switchstate"
               },
               {
-                "title": "Invoke State",
-                "$ref": "#/definitions/invokestate"
+                "title": "SubFlow State",
+                "$ref": "#/definitions/subflowstate"
               }
             ]
           }
@@ -512,9 +512,9 @@
         }
       }
     },
-    "invokestate": {
+    "subflowstate": {
       "type": "object",
-      "description": "Defines a local workflow to be executed",
+      "description": "Defines a sub-workflow to be executed",
       "properties": {
         "name": {
           "type": "string",
@@ -522,7 +522,7 @@
         },
         "type": {
           "type" : "string",
-          "enum": ["INVOKE"],
+          "enum": ["SUBFLOW"],
           "description": "State type"
         },
         "end": {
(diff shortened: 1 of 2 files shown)


workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -452,6 +460,47 @@
...
+        "filter": {
+          "$ref": "#/definitions/filter"
+        },
+        "next-state": {
+          "type": "string",
+          "description": "Specifies the name of the next state to transition to after local workflow has completed execution."
+        }
+      },
+      "required": ["name", "type", "next-state", "workflow-id"]
+    },
```
cathyhongzhang: Great! I was thinking about adding something like this.  Thank you! 
Just wondering if we need workflow-version since the workflow-id already uniquely identifies the workflow (My understanding of the version is to identify a specific workflow spec if it has several evolving updates)

workflow/spec/spec.md
```diff
@@ -911,6 +913,86 @@ The elements of the output array need not be of the same type.
...
+### <img src="media/state-icon-small.png" with="30px" height="26px"/>Invoke State
+
+| Parameter | Description | Type | Required |
+| --- | --- | --- | --- |
+| name |State name | string | yes | 
+| type |State type | string | yes | 
+| end |If this state and end state | boolean | no |
+| wait-for-completion |If workflow execution must wait for local workflow to finish before continuing | boolean | yes |
+| workflow-id |Local workflow unique id | boolean | no |
+| workflow-version |Local workflow version | boolean | no |
```
cathyhongzhang: ditto

workflow/spec/spec.md
```diff
@@ -911,6 +913,86 @@ The elements of the output array need not be of the same type.
...
+            "type": "boolean",
+            "default": false,
+            "description": "Workflow execution must wait for local workflow to finish before continuing."
+        },
+        "workflow-id": {
+            "type": "string",
+            "description": "Local workflow unique id."
+        },
+        "workflow-version": {
+            "type": "string",
```
cathyhongzhang: ditto

workflow/spec/spec.md
cathyhongzhang: ditto

workflow/spec/spec.md
```diff
@@ -911,6 +913,86 @@ The elements of the output array need not be of the same type.
...
+        }
+    },
+    "required": ["name", "type", "next-state", "workflow-id"]
+}
+```
+
+</details>
+
+It is often the case that you want to group your workflows into small, **reusable** logical units that perform certain needed functionality.
+Even tho you can use the Event state to call an externally deployed services (via function), at times
```
cathyhongzhang: change tho to though

workflow/spec/spec.md
```diff
@@ -911,6 +913,86 @@ The elements of the output array need not be of the same type.
...
+Another use of this state is within [branches](#parallel-state-branch) of the [Parallel State](#Parallel-State). Instead of having to define all states
+in each branch, you could separate the branch states into individual included workflows and call the Invoke state
+as a single state in each.
+
+The called included workflow must have a defined start and end states. 
+The wait-for-completion property defines if the Invoke state should wait until execution of the included workflow
+is completed or not. 
+
+Each invoked workflow receives a copy of the Invoke state's input data.
+Included workflows executed are not allowed to edit the parent workflows' data.
```
cathyhongzhang: why do we have this restriction? There might be cases in which the purpose is to have the sub-workflow modify the parent workflow's data.

workflow/spec/schema/serverless-workflow-schema-01.json
cathyhongzhang: Note sure if "invoke state" is the best name. Something like "subworkflow state" is more self-explanatory, but the naming is too long. Would you like to provide a better name?

tsurdilo: ok will update both.

workflow/spec/spec.md
tsurdilo: Could we add a flag like "independent": true/false so users can choose? If true then the subflow would maintain its own json context object. If false it would have access to the parent workflow context. wdyt?

cathyhongzhang: If wait-for-completion is false, the parent workflow will not wait for the subflow result. In this case we should not allow subflow to modify the main flow's data since we do not want timing to decide the data on which the main workflow operates. If wait-for-completion is true, then we should not restrict the subflow to modify the data. What do you think?

tsurdilo: ok. will update.

tsurdilo:
@cathyhongzhang changed state name to "SubFlow", removed workflow-version property. Updated the information passing info (depending on wait-for-completion) property.

tsurdilo:
rebased and updated props to use camelcase.

tsurdilo:
rebased again

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -504,6 +512,43 @@
         }
       }
     },
+    "subflowstate": {
+      "type": "object",
+      "description": "Defines a sub-workflow to be executed",
+      "properties": {
+        "name": {
```
ruromero: I think you're missing the new `id` property

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -504,6 +512,43 @@
...
+          "type": "string",
+          "description": "Unique name of the state"
```
ruromero: If the `id` property is added, this is no longer _Unique_

workflow/spec/spec.md
```diff
@@ -1006,6 +1008,81 @@ The elements of the output array need not be of the same type.
...
+    },
+    "required": ["name", "type", "nextState", "workflowId"]
+}
+```
+
+</details>
+
+It is often the case that you want to group your workflows into small, **reusable** logical units that perform certain needed functionality.
+Even tho you can use the Event state to call an externally deployed services (via function), at times
+there is a need to include/nest another serverless workflow (from classpath/local file system etc). In that case you would use the SubFlow State.
```
ruromero: I would add that the concept of local workflow may vary depending on the implementation. It can also be a Kubernetes namespace, for example.

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: updated

workflow/spec/spec.md

workflow/spec/spec.md

workflow/spec/spec.md

workflow/spec/spec.md

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: updated

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: updated

workflow/spec/spec.md
tsurdilo: good point, added "depending on the implementation logic" to note that in doc. wdyt?

PR #101, opened 2019-11-08
Element Metadata

tsurdilo:
This PR adds "Element metadata" to each element of the workflow definitions. 

Why this is needed - Just like we allow users to add custom extensions, we need to add the ability for users and implementors to enrich the core definition of workflow elements with their own/custom information. This information should not affect process execution however can be very helpful for implementations/users/tooling.

Element metadata description (as included in this pr):
*************

## Element Metadata

Element metadata enables you to enrich elements of the serverless workflow
with information above and beyond that which is defined its core definition. 
This is done via the "metaData" property (included in each workflow element) and which has the main goal of allowing users to add additional 
descriptions of the element it belongs to.

Metadata information should not affect workflow execution. Implementations have the choice to act upon metadata info
or ignore it. 
Metadata can be useful for things like documentation, content management, authentication, translation, logging, etc (basically things above and beyond core workflow execution).

The metadata property can be seen as a map where keys are of type string and values have the object type. 

Here is an example of metadata property defined in the main workflow definition. Note that 
all metaData parameters are just examples, you can fully define your own.

```json
{
  "id": "workflow-uuid",
  "name": "workflow-name",
  "version": "1.0",
  "metaData": {
    "loginfo": {
      "enabled": true,
      "level": "INFO"
    },
    "authInfo": {
      "organization": "Org name",
      "group": "Group name",
      "type": "Auth type",
      "token": "Token",
      "keys": {
        "kty": "RSA",
        ...
      }
      ....
    }
  },
  "states": [
    .....
  ]
}
```


salaboy:
@tsurdilo totally agree with this.. each element should have metadata (another bag of properties that implementors can use to add more context). 
I started to think as it is handled in Kubernetes where we have metadata and it is divided into annotations and (labels and selectors).. maybe we can follow the same approach

https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

salaboy:
@tsurdilo also.. just a detail.. instead of `metaData` I would rather use `metadata` not sure why.. I am already so used to see it without camelcase that I think that it will make sense to keep it like that.. unless you feel strongly about the other approach.

tsurdilo:
@salaboy good points. Will look at the kubernetes metadata approach. Will update to "metadata".

tsurdilo:
@salaboy i read the docs you linked but I don't really understand the difference. It defines:
Labels:
```
Labels are key/value pairs that are attached to objects, such as pods. Labels are intended to be used to specify identifying attributes of objects that are meaningful and relevant to users, but do not directly imply semantics to the core system. 
```

Annotations:
```
You can use Kubernetes annotations to attach arbitrary non-identifying metadata to objects. Clients such as tools and libraries can retrieve this metadata.
```

Sounds same to me honestly :) Maybe I'm wrong tho so asking. 
What I do see is that maybe we want to limit metadata to key/value pairs both type string, where I defined it here as key string value type object. That I can change if we think it's better.

salaboy:
@tsurdilo yes... it is a subtle difference but an important one on the implementation side. I think that labels are indexed and annotations are not. Mostly because labels are used to retrieve heterogeneous resources (all the things that have these labels set) and annotations are more used for extensions specific scenarios. So basically the metadata is divided in core things such as labels and then annotations which are for others to add.

tsurdilo:
@salaboy thanks for explaining. I updated the metadata definition. It now incldes "tags" and "notes" elements which are key/value pairs. Pr includes "Element Metadata" section which describes them and provides an example. Let me know if this is ok now.

workflow/spec/spec.md
```diff
@@ -1141,6 +1234,51 @@ logging, simulation, debugging, tracing, etc.
 
 You can find more info and examples of element extensions [here](spec-extending.md).
 
+## Element Metadata
+
+Element metadata enables you to enrich elements of the serverless workflow
+with information above and beyond that which is defined its core definition.
```
cathyhongzhang: typo. change to "beyond what are defined in its core definition"

workflow/spec/spec.md
```diff
@@ -1141,6 +1234,51 @@ logging, simulation, debugging, tracing, etc.
...
+descriptions of the element it belongs to.
+
+Metadata can be grouped into two categories 
+
+- **tags**:  key/value pairs (type string) which describe
+element attributes relevant to users. Tags do not affect workflow execution. Example tags could include "protocol", "authentication", "release", "system", etc.
+Each tags key must be unique. Tags can be used for indexing, searching, and organizing workflow elements.
+- **notes**: key/value pairs (type string) which can include arbitrary information (not related to the element they are attached to). Example notes could include "gitinfo", "team", "docsurl", "buildinfo", etc.
+Each notes key must be unique. Notes can be used by clients (tools, libraries) for display or content management purposes.
+
```
cathyhongzhang: When I read the explanation of tags and notes, it is not very clear to me which ones I should put in tags and which I should put in notes. The boundary is not clearly defined. I am wondering if two categories are enough or the best. Maybe we just need one field "metadata" for an element.

tsurdilo: information in tags is supposed to be related to the element it belongs to. notes contains information that does not describe something directly related to the element it belongs to, but provides important information for things like tooling etc. 

Maybe I can work on making this documentation clearer. will try

workflow/spec/spec.md
tsurdilo: fixed

workflow/spec/spec.md
tsurdilo: @salaboy i also think its not very clear what information to put in tags/notes or labels/annotations. Maybe we could make metadata a key/value bucket without this specification for now and if we see it's needed in the future we can add? WDYT?

tsurdilo:
i will remove the two metadata buckets and allow users to add key/value pairs to it as they wish for now. if we find a need to add specific labeled buckets in the future we might revisit that at that point.

tsurdilo:
@cathyhongzhang could you please review again, updated to remove the tags/notes buckets. metadata is map of key/value pairs (type string).

tsurdilo:
Putting this PR "On Hold" as we should currently focus on core specification details. This can be revisited at a later point.

cathyhongzhang:
Hi Tihomir, 
I am a little confused. Would you like to make this on hold or would you like me to continue the review? Thanks.

tsurdilo:
@cathyhongzhang I put this "on hold" by mistake (sorry). Yes its currently only "Review Needed". Definitely need your review after the last updates based on your comments please .

cathyhongzhang:
sure, np. reviewing it now.

workflow/spec/spec.md
```diff
@@ -1141,6 +1219,43 @@ logging, simulation, debugging, tracing, etc.
...
+    "author": "author name",
+    "team": "team name",
+    ...
+  },
+  "states": [
+    ...
+  ]
+}
+
+```
```
cathyhongzhang: Since they could be some arbitrary information, I am thinking whether it is better to add this metadata into the extension bag (maybe define some explicit fields for tooling and debugging usage). From my past experience, it is better to limit the definition of a spec's core structs/elements to a minimum. We would not like to see people throwing arbitrary information in a key struct. CloudEvents went through this exercise and eventually only added key structs/attributes into the spec. WDYT?

cathyhongzhang: If we can not clearly define a struct/element/field, then people could have different interpretation in the future, which might lead to inconsistent or even conflict use of the spec .

tsurdilo: @cathyhongzhang metadata is very similar (and more restricted) than the "data" element of cloudevents. There the data element allows values of type object where we allow for metadata key/value pairs of type string.  Users can put custom information in this "data" element to enhance the event description, similarly here we have the metadata element which can add this type of info to each element of the workflow definition.

cathyhongzhang: I understand that there is some similarity to the data/payload of a CloudEvents. But I still feel the metadata here is kind of different from the data/payload of CloudEvents. Here the metadata is defined as key/value pairs (type string) which can include arbitrary information (not related to the element they are attached to). But For Cloud Event, payload is not something that is arbitrarily defined, and it is a natural part of that event. For example, the payload for a motion event would be an image capturing the motion. This definition of metadata is not clear and I am afraid that people will throw arbitrary information to it. If so, what is the problem with putting those information to the extension bag?

tsurdilo: I don't fully understand why "people throwing arbitrary information to it" is a bad thing or why it is considered dangerous? The amount of benefits it brings far overweights the possibility of someone misusing it, and even then since we define that metadata information should not affect workflow execution it really only has benefits regardless how people use it (or use it).

Regarding "Payload is not something that is arbitrarily defined, and it is a natural part of that event", that may be the logical thinking but thats not how it is defined. Who are we to say what is "natural" information for any/all users that model serverless workflow and what is "unnatural"? I could include my uncles friends cousins car information inside this image capturing payload and if this makes sense to my business and my business logic who is to say that it is wrong?

The reason to not put metadata inside "extensions" is that metadata information CAN be related and is close-to the element it belongs to. Extension information as we defined it before deals more with cross-cutting concerns involving the entire workflow model, and not individual nodes that make up the workflow model. Hope this helps.

tsurdilo:
@cathyhongzhang can we please make a decision on this. we have prs sitting around for weeks now. its starting to look like this project is not active again.

tsurdilo:
@salaboy can you please comment on this pr? Thank you!

tsurdilo:

tsurdilo:
rebased against master

salaboy:
@tsurdilo this looks good to me.. I know that metadata is a fundamental concept to have in these kind of languages
/LGTM

tsurdilo:
@mswiderski ok i think @cathyhongzhang had the same concerns. I will rethink this a little bit to see if already added extension element is enough and come back with examples.

cathyhongzhang:
@tsurdilo Just FYI. The day you requested comment/decision is our Thanksgiving holiday (11/28~11/29) :-)

cathyhongzhang:
@tsurdilo You mentioned that you will rethink this. Just FYI. I will be on vacation next week and will review the changes when I am back to office.

cathyhongzhang:
Are you going to update this? If not, could you close it or put it in another status?

tsurdilo:
@cathyhongzhang we have myself and @salaboy thinking this change is good to have an yourself and @mswiderski made good comments regarding using just "extensions". So yeah I'll change the status to "On Hold" for now.

tsurdilo:
After thinking about this for a while I still think that having metadata is useful for implementations. Will work on an update and either add to this pr or create a new one..depending how hard it is to rebase this pr after all the other updates to spec.

salaboy:
@tsurdilo is this on hold, should I review it as it is?

tsurdilo:
@salaboy I'm working on a new one . its too hard to merge this one after it has been sitting around for a while. will close this one soon.

tsurdilo:
will close this pr and create a new one for metadata. this one has been sitting around for too long and rebasing it at this point is just too hard.

PR #103, opened 2019-11-12
State Looping behavior

tsurdilo:
This pr adds looping behavior to each state.
Instead of having a single state which allows looping over some data collection, in our case it makes more sense to add looping to each state defined. 
Looping happens over each element of a input collection. In addition a completion condition can be set which if true stops looping. 
If a state defines data results, each iteration/loop adds the states results to the output collection.
After each iteration/loop users can define a time delay to wait before next iteration/loop can start.

### DIFF
--- a/workflow/spec/spec.md
+++ b/workflow/spec/spec.md
@@ -1253,10 +1253,10 @@ Filters are used for data flow through the workflow. This is described in detail
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| inputCollection |Selects a collection of the states inputPath | string | no |
+| inputCollection |Selects a collection of the states inputPath | string | yes |
 | outputCollection |Selects a collection of the states outputPath | string | no |
 | completionCondition |Boolean expression (evaluated after each iteration) that controls the loop. If this expression is true looping stops even if not all collection elements are processesd | string | no |
-| timeDelay |Amount of time (ISO 8601 format) to wait between each loop | integer | yes |
+| timeDelay |Amount of time (ISO 8601 format) to wait between each loop | string | no |
 
 <details><summary><strong>Click to view JSON Schema</strong></summary>
 
--- a/workflow/spec/spec.md
+++ b/workflow/spec/spec.md
@@ -270,6 +270,7 @@ We will start defining each individual state:
 | end |Is this state an end state | boolean | no |
 | [events](#eventstate-eventdef) |Array of event | array | yes |
 | [filter](#Filter-Definition) |State data filter | object | yes |
+| [loop](#Loop-Definition) |State loop information | object | yes |
  
 <details><summary><strong>Click to view JSON Schema</strong></summary>
 <p>
@@ -308,7 +309,10 @@ We will start defining each individual state:
         },
         "filter": {
           "$ref": "#/definitions/filter"
-        }
+        },
+        "loop": {
+          "$ref": "#/definitions/loop"
+        } 
     },
     "required": ["name", "type", "events"]
 }
@@ -554,6 +558,7 @@ as well as define parameters (key/value pairs).
 | actionMode |Should actions be executed sequentially or in parallel | string | yes |
 | [actions](#Action-Definition) |Array of actions | array | yes |
 | [filter](#Filter-Definition) |State data filter | object | yes |
+| [loop](#Loop-Definition) |State loop information | object | yes |
 | [nextState](#Transitions) |State to transition to after all the actions have been successfully executed | string | yes (if end is set to false) |
 
 <details><summary><strong>Click to view JSON Schema</strong></summary>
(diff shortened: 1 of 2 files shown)


tsurdilo:
could we get a review on this please? thanks.

workflow/spec/spec.md
```diff
@@ -1127,6 +1151,86 @@ If this property is sete to false, data access to parent's workflow should not b
...
+  "required": ["inputCollection"]
+}
+```
+
+</details>
+
+Loop definition allows states to be executed in a loop until a completion condition becomes true. 
+
+Looping happens over each element of the inputCollection parameter. This parameter must evaluate to a collection
+of the states input data.
```
cathyhongzhang: I am trying to understand how this will be used.  
1. It is not clear to me what is meant by "a collection of the states inputPath"? Does the inputPath refer to the input data to this state? What does a collection mean? 
2. Does the workflow first execute this state, then evaluate the CompletionCondition? But the evaluation the example gives is InputCollection, which seems to determine whether to end the loop or not. 
3. It will be helpful for us to review if you can clarify what are the purpose and usage of InputCollection and OutputCollection? 

Thanks!

tsurdilo: 1. A collection is grouping of data (in this case an array or a list etc....). Yes, inputCollection parameter defines a path to a collection (lets say array) defined in the states data input.

2. If you are asking about implementation details - if workflow model defines a state with a "loop" parameter, first the inputCollection path is evaluated. For each element of the input collection data, the state will be executed. After each iteration, also the completionCondition is checked which can break this loop if condition is met. Hope this helps you understand

3. Currently if you look at a state definition, a state has data input and output. What you are asking here is the exact same concept, except with looping involved you have a collection of data to loop over (inputCollection) and each state execution (iteration/loop) can produce an output which then is added to the outputCollection.

tsurdilo:

tsurdilo:
rebased against master

workflow/spec/spec.md
```diff
@@ -1186,6 +1209,86 @@ If this property is sete to false, data access to parent's workflow should not b
 
 Filters are used for data flow through the workflow. This is described in detail in the [Information Passing](#Information-Passing) section.
 
+### Loop Definition
+
+| Parameter | Description | Type | Required |
+| --- | --- | --- | --- |
+| inputCollection |Selects a collection of the states inputPath | string | no |
```
mswiderski: shouldn't the input collection be required otherwise there is nothing to loop over?

workflow/spec/spec.md
```diff
@@ -1186,6 +1209,86 @@ If this property is sete to false, data access to parent's workflow should not b
...
+| outputCollection |Selects a collection of the states outputPath | string | no |
+| completionCondition |Boolean expression (evaluated after each iteration) that controls the loop. If this expression is true looping stops even if not all collection elements are processesd | string | no |
+| timeDelay |Amount of time (ISO 8601 format) to wait between each loop | integer | yes |
```
mswiderski: why this one is required? I'd find it more intuitive to make it optional and only applied when given,wdyt?

salaboy: I kinda agree.. time delay between executions should not be mandatory.

workflow/spec/spec.md
salaboy: To give an example of how this can be used.. imagine that you have a CloudEvent that comes with a list of Documents as payload. You might want to validate each document individually by calling a function. 
With the loop state you can use the list of documents as input and then define the actions for each element, instead of the whole group.

salaboy:
@tsurdilo besides the comment from @mswiderski this looks good to me to merge
/LGTM

workflow/spec/spec.md
tsurdilo: agreed. will update

workflow/spec/spec.md
tsurdilo: same :) will update

tsurdilo:
updated based on @mswiderski comments

tsurdilo:

workflow/spec/spec.md
cathyhongzhang: I think this is a good feature. What I am thinking is whether we can reuse the filter defined in the spec. 
If  inputCollection defines a path to the state's input data, then we may reuse the filter since the filter is also a path to state's input data. Same goes for the outputCollection. But it seems that your  inputCollection carries more semantics than just a path filter since you mentioned that the inputCollection path is evaluated. Does inputCollection include the evaluation definition? If so, what kind of evaluation does it have (e.g. equal or larger than, etc. or what)?

tsurdilo:
rebased against master

workflow/spec/spec.md
tsurdilo: it just says it must be a collection type. same for outputCollection

tsurdilo: its not same as filters currently, as it contains completion condition and time delay. I will look into this is future pr tho to see what all things we currently have can be combined and such. thanks

PR #105, opened 2019-11-12
[WIP] Proposed examples and terminology

ruromero:
I wanted to create this PR to provide some real life examples that will help us better define the scope and also identify what might be missing in the current version of the spec. My intention with this PR is not to have it accepted or rejected but to start an open discussion and work around these examples (and maybe provide more). I obviously took some fields from AsyncAPI which I find a good match for what this spec is aiming for but I didn't know how to leverage it for the shared aspects.

These are the main points to be discussed. Please take my examples as exactly that, examples. I would like to work with anyone interested in taking whatever is considered useful.

**CloudEvents**
They are the most important element in a Serverless Workflow this is what is used to trigger workflows and to interact with functions. CloudEvents payload should be in JSON format because the spec uses JSONPath for mapping/filtering/conditionals/correlationTokens

* Type definitions: Having type definitions will enable validation but also properly use JSON Path. External schemas can be used, it's just a matter of using remote url in the `$ref`. CloudEvent Attributes are also used to allow more flexible event mappings.
* Servers definitions: In order for the serverless workflow to act as a producer/consumer of cloudEvents it should be able to define where to receive events from and where to send events to. This will allow the workflow to interact with cloudEvent-enabled services and brokers (e.g. Knative or Kafka)

**Data flow**
In the current version of the spec it is described that data will be passed from one state or function to the next as is but, does that include the `ce-id`, `ce-type`, `ce-source` ? Where are these defined? That should be covered by the events definition and the `eventSource` attribute. This attribute sets the source to each event emitted from the workflow. The ID should be generated and type is defined in the events section. For additional attributes, it should be possible to be done through filters.
* Filtering: Or mapping, is how the message is transformed before or after being processed by a state/event/function. Currently it only allows for the case of the input to provide a jsonPath as the payload to use but what if we need to combine different fields or provide CloudEvent attributes? Same happens for the output and result path. See my proposal around filtering.

**Event states**
Event states should be the only entry point for a Serverless Workflow that is event-driven. When receiving the event, mappings (i.e. filters) can be applied to take from the attributes and payload so that messages can be enriched/transformed.
I don't think it is needed to have complex event filtering. The type/source of the cloudEvent should be enough. If not, let's find an example.
I see very easy to follow the fact that events are defined and identified with a unique name that will be used in the event states. Event of type A is received from server X -> send to state M / Event of type B is received from server Y -> send to state N

**Function calls**
Again, we're focusing on CloudEvents so we only send/receive this type of events. That means a function only needs to specify the event type and where to send it. Then we decide how to proceed once the event is sent.
There are 2 possible cases:
* Fire and forget. It is a notification event, the workflow doesn't care and the message is not modified
* Async call. The workflow sends the event and waits for a callback event to be received. The event will be matched using a correlationToken defined at this point because in some other function call. It can be a different one. See the example of the `username` in the user_registration workflow.

Besides, I don't think having an OPERATION state with all the functions inside is really needed. A single FUNCTION state should suffice. When a workflow grows in complexity it should be possible to split it into multiple workflows.

See the `waitFor` example where the name of the event is defined and also the server where it should come from. Also there is a correlationToken as described above together with a timeout and a retry configuration.

**SWITCH and PARALLEL states**
I haven't covered these examples yet but I think this PR already covers a big bunch of interesting topics.

### DIFF
--- a/workflow/spec/spec.md
+++ b/workflow/spec/spec.md
@@ -5,14 +5,14 @@
 Serverless applications are becoming increasingly complex. Nowdays they have to coordinate, manage, and define
 the execution order (steps) for countless functions triggered by as many events.
 
-When we are dealing with large number of functions, managing their execution is not a simpler task. 
-For example we have to coordinate functions and event triggers, orchestrate function execution (sequential,  parallel, 
+When we are dealing with large number of functions, managing their execution is not a simpler task.
+For example we have to coordinate functions and event triggers, orchestrate function execution (sequential,  parallel,
 in branches depending on different event triggers), etc.
 
 Workflows have become key components of serverless applications as they excel at orchestration and coordination
-of their functional flow. 
+of their functional flow.
 
-The goal of the Serverless Workflow sub-group is to come up with a standard way for users to specify their serverless application workflow, as well as help facilitate 
+The goal of the Serverless Workflow sub-group is to come up with a standard way for users to specify their serverless application workflow, as well as help facilitate
 portability of serverless applications across different vendor platforms.
 
 Serverless Workflow is a vendor-neutral and portable specification which meets these goals.
@@ -23,29 +23,31 @@ This document is a working draft.
 
 ## Table of Contents
 
-- [Introduction](#Introduction)
-- [Use Cases](#Use-Cases)
+- [Overview](#Overview)
+  - [Functional scope](#Functional-scope)
+- [Notation and terminology](#Notation-and-terminology)
+- [Use cases](#Use-cases)
 - [Specification Details](#Specification-Details)
-    - [Workflow Model](#Workflow-Model)
-    - [Workflow Definition](#Workflow-Definition)
+  - [Workflow Model](#Workflow-Model)
+  - [Workflow Definition](#Workflow-Definition)
 - [Extending](#Extending)
 - [Examples](#Examples)
 - [Reference](#Reference)
 
-
-## Introduction
+## Overview
 
 Serverless Workflow can be used to:
 
-* **Orchestrate serverless application logic**: serverless applications are typicall event-driven and can be 
-very hard to manage. Serverless Workflow groups the application events and functions into a coherent unit and 
+- **Orchestrate serverless application logic**: Serverless applications are typicall event-driven and can be
+very hard to manage. Serverless Workflow groups the application events and functions into a coherent unit and
 simplifies orchestration of the app logic.
-* **Define and coordinate application control flow**: allow the users to define the execution/operation
+- **Define and coordinate application control flow**: allow the users to define the execution/operation
 control flow and how/which functions are to be invoked on arrival of events.
-* **Define and manage application data flow**: allows the users to define how data is passed and filtered from incoming events to states, 
-rom states to functions, from one function to another function, and from one state to another state.
+- **Define and manage application data flow**: allows the users to define how data is passed and filtered from incoming events to states,
+from states to functions, from one function to another function, and from one state to another state.
 
 ### Functional Scope
+
 Serverless Workflow allows users to:
 
 1. Define and orchestrate steps/states involved in a serverless application.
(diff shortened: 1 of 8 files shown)


tsurdilo:
@ruromero super useful examples that will help move changes forward I think.
Can I just ask - some of changes here for spec.md could be applied currently? Would it make sense to create a separate PR for them so they can be included independently of the examples?

workflow/spec/spec.md
```diff
@@ -1204,19 +1245,13 @@ Each Filter has three kinds of path filters
   - Specify output data of State or Action as JSONPath
   - Default value is '\$'
 
-<p align="center">
-<img src="media/filter-sequential.png" with="480px" height="270px" alt="Sequential FilterDiagram"/>
```
tsurdilo: I made these changes to keep images smaller and same size in the main spec doc. Users can view them in full size by clicking on them. Are they too small now to revert to the full size view from the start?

ruromero: Nope, I was just trying to use as much markdown as possible as the other changes in the doc

ruromero:
> I just ask - some of changes here for spec.md could be applied currently? Would it make sense to create a separate PR for them so they can be included independently of the examples?

If you feel they are fine as they are now or with little changes, then yes. If they're part of the discussion then let's keep them. WDYT?

workflow/spec/spec.md
```diff
@@ -59,38 +61,99 @@ Serverless Workflow allows users to:
...
+#### Event
+
+An event is an instance of a CloudEvent defined in the workflow. An event consists of the payload and the headers.
+Headers can be used for event filtering and data transformation at the
+event reception. The event payload MUST be `application/json` because events are defined using Json Schema
+and JSONPath is used for message  transformations.
+
+#### State
+
+States define the logic of the workflow. A state can react to events, call functions, make decisions and create branches. The state MUST use the message resulting from the previous transition and SHALL be replaced by the output of the processing of the current state. The resulting message MUST be passed to the next transition.
```
tsurdilo: I think states on their own are juts building blocks. Maybe change to sentence saying like they are on their own building blocks but can be used to model business requirements / logical workflow definitions / something like that. 

Also if we change things here same type of description should be given in the State Definition section (and maybe link those two together). Same with other definitions here (links)

workflow/spec/spec.md
```diff
@@ -59,38 +61,99 @@ Serverless Workflow allows users to:
...
+#### Workflow
+
+A Serverless Workflow is a definition of an event orchestration which MUST be repeatable.
+A workflow receives one or more events that will be used as the input for the defined states.
+States use the current message to call serverless functions and transform the message.
+The data will transition from one state to the next until no more states are defined.
+
+#### Message
+
+A message is the object that will flow throughout each transition. The message MAY or MAY NOT be transformed by a state or its filters. The message MUST be created from the payload of an incoming event or replaced from the payload of a response event, from a function call.
```
tsurdilo: Not sure we should call this a "message" imo implies it may have to do w/ messaging protocols etc. We need to also discuss this whole topic imo at some point as far as data containment/flow, global data vs private etc. Probably best to use data input/output atm. But I think this is good to have here as it can drive updates to the information passing section

workflow/spec/spec.md
```diff
@@ -59,38 +61,99 @@ Serverless Workflow allows users to:
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", "MAY NOT" and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+### Terminology
+
+This specification defines the following terms:
+
+#### Workflow
+
+A Serverless Workflow is a definition of an event orchestration which MUST be repeatable.
```
tsurdilo: I would avoid use of "orchestration" only here as in Introduction section we define that workflow can be used for more than just that.
Also imo would avoid "repeatable" for now as that implies lots of other things that currently we haven't imo fully defined yet (just saying i dont think we are there yet).

workflow/spec/spec.md
```diff
@@ -59,38 +61,99 @@ Serverless Workflow allows users to:
...
+A workflow receives one or more events that will be used as the input for the defined states.
```
tsurdilo: There needs to be discussion if workflow can start with a pre-defined state first imo (future discussion). 
This also may imply that data must come from events defined in triggers which then forces the use of event state as the starting point of any workflow imo.

workflow/spec/spec.md
ruromero: If these changes are accepted, they will have to be propagated to the rest of the spec (text and schema). Let's keep this conversation opened til we agree on the wording for states. I don't have anything against your suggestion.

workflow/spec/spec.md
```diff
@@ -59,38 +61,99 @@ Serverless Workflow allows users to:
...
+States use the current message to call serverless functions and transform the message.
+The data will transition from one state to the next until no more states are defined.
```
tsurdilo: Maybe change until workflow reaches and completes an end state, as there could be multiple paths and some may not cover all states defined.

ruromero: I have based these definitions on the examples provided which I consider easy to follow and understand. There I haven't found the need to add neither the `startAt` nor the `end` as for the case of the `gardening` example there are multiple entry points and paths.

workflow/spec/spec.md
```diff
@@ -59,38 +61,99 @@ Serverless Workflow allows users to:
...
+
+#### Message
+
+A message is the object that will flow throughout each transition. The message MAY or MAY NOT be transformed by a state or its filters. The message MUST be created from the payload of an incoming event or replaced from the payload of a response event, from a function call.
+
+#### Function
+
+A (serverless) function is a remote service that is invoked in the workflow using the current message. The serverless function MUST support CloudEvents and the content type of the event will be `application/json`.
```
tsurdilo: Maybe we should just say invokable service here instead of remote? 
Also would not insist that current message or data or whatever we want to call it must be involved as to assume all these services will require data to be invoked. wdyt?

workflow/spec/spec.md
```diff
@@ -59,38 +61,99 @@ Serverless Workflow allows users to:
...
+The serverless function MAY or MAY NOT reply synchronously. In case a synchronous reply is received and such response is a CloudEvent it MAY be used as output of the current state.
```
tsurdilo: This section is a little confusing to me, maybe show some examples? Also I think you define the transition object as "message" earlier and here mention "state output".

workflow/spec/spec.md
```diff
@@ -59,38 +61,99 @@ Serverless Workflow allows users to:
...
+The function call MAY wait for an asynchronous event to be considered finished. The function SHALL define how to correlate both, the sent and the received events.
```
tsurdilo: We might want to link here to attributes like actionMode and waitForCompletion etc where applicable to back up the may/must/shall statements

workflow/spec/spec.md
ruromero: Yes, that's my point. I consider Serverless Workflows should be defined around CloudEvents. So a workflow should react to events. There can also be an event without payload to notify that the workflow should be started. How should then a user in a cloud environment trigger a workflow otherwise?

workflow/spec/spec.md
```diff
@@ -59,38 +61,99 @@ Serverless Workflow allows users to:
...
+A reply, either synchronous or asynchronous is OPTIONAL, in case no reply is needed the message MUST NOT be modified.
+
+#### Event
+
+An event is an instance of a CloudEvent defined in the workflow. An event consists of the payload and the headers.
```
tsurdilo: Maybe link to trigger definitions from here? wdyt

workflow/spec/spec.md
tsurdilo: yeah agree. all good points that need to be discussed!

workflow/spec/spec.md
tsurdilo: ok will look more into examples then. my bad :)

workflow/spec/spec.md
cathyhongzhang: I think a workflow's scope will be larger than an event orchestration although event is an important part of it. Some steps of the workflow do not involve any events. Could you clarify what the repeatable means here?

cathyhongzhang:
@ruromero These are interesting points for a f2f discussion. I would like to better understand your thoughts. How about we discuss them in the Kubecon? Thanks

workflow/spec/spec.md
cathyhongzhang: I also thought it should be event triggered. But we should allow the user to trigger the workflow through API or CLI, right? In some scenarios, the admin might want to start the workflow manually unless we say the API/CLI triggers are also events.

workflow/spec/spec.md
cathyhongzhang: It is not always true that data will transit from one state to the next state.

ruromero:
> @ruromero These are interesting points for a f2f discussion. I would like to better understand your thoughts. How about we discuss them in the Kubecon? Thanks

That sounds great. I wanted to share all this first to give you all some time to discuss and also give others the chance to chime in.

workflow/spec/spec.md
tsurdilo: yes definitely should be: 
- start workflow from api / cli / rest call 
- start workflow in a known state (from persisted data)
- be able to predefine workflow data (for testing for example)

so events are big part of the spec but do not define it. serverless workflow should have value if no events are present as well (simple orchestration scenario for example)

workflow/spec/spec.md
ruromero: Can you elaborate, please?

workflow/spec/spec.md
ruromero: Repeatable means that the workflow doesn't keep any state. A workflow will behave the same for any input, regardless of its logic.
My point regarding events is that any function call should be done through CloudEvents. This should be IMO the only communication mechanism. Why would you need a workflow if you don't interact with other systems?

workflow/spec/spec.md
ruromero: Agree on avoid using _message_
Information passing and transformation is a key point. I agree.

workflow/spec/spec.md
ruromero: A rest call can also be a cloudEvent so that works for api/cli/rest.
To start a workflow in a known state from persisted data should be up to the implementation but a workflow implementation that does not persist data and never stops (hypothetically) should work the same as one that does persist data.
If you want to use predefined workflow data for testing, use mocks. Testing should not condition the spec.

Same applies to my other comment. Orchestration of what? Orchestration of services through a common mechanism, i.e. CloudEvents. I don't think this spec should try, at least from the beginning to consider anything else than CloudEvents. IMO they allow an homogeneous and consistent Serverless communication.

workflow/spec/spec.md
tsurdilo: Maybe we are using different terminology here? Not keeping any state is "stateless". Repeatable should mean that provided the same input multiple times the workflow should provide the same (predicted) results.

Workflows should not be limited to being stateless as we define workflow itself to define and maintain state. Also we describe one of the purposes of the spec to be coordinate and manage application data, so stateless does not fit there at all. 

I don't understand statement: -- any function call should be done through cloudevents -- my understanding is that cloudevents describes events and event data, not define how functions /apps are called/invoked so not sure I understand that statement and last question.

workflow/spec/spec.md
tsurdilo: Orchestration of functions/apps, which does not depend on event existence (but can be driven by it). 
It already defines events (triggers) to be cloudevents. As far as invocation of functions/apps we don't  limit that.

ruromero: If majority agrees that serverless workflows should aim for any type of application integration I will accept it. However, my personal opinion remains the same. I would prefer function communication to be only driven by CloudEvents for design simplicity and because I find it better suited for a Serverless environment. This is what I understand when reading about functions in the [serverless whitepaper](https://github.com/cncf/wg-serverless/tree/master/whitepapers/serverless-overview#function-requirements) but my interpretation might not be accurate.

For traditional application integration there are other specs and implementations broadly adopted and I don't think we should provide a different solution for the same problem but provide something more specific for serverless.
Again, this is a personal opinion and I will accept the opinion of the majority.

workflow/spec/spec.md
ruromero: Your definition of _repeatable_ is much clearer, I wasn't concise enough.

CloudEvents also define different bindings that can be used when defining the different events/servers in the workflow. An event A will be of type M and use a binding X. In the _servers_ section, included in the examples, it can be defined how to interact with each server (url and protocol), but as any example it's just an idea where more complexity can be added if needed(auth/headers/load balancing/...).

tsurdilo:
I think the comments so far have been about the text, but still need to look at the json examples which introduce quite big changes to the current definitions that we have. It would help if changes you introduce such as new parameters, new states, dynamic inclusion via $ref etc (I am still going through them) would have explanation on like what they change and why so its better understandable for everyone.

ruromero:
The main topics were discussed during the f2f and is better to close this PR and use the comments and feedback gathered in the next iterations.
Thanks for your comments so far.

PR #109, opened 2019-11-14
JSON Schema fixes

tsurdilo:
This fixes issues like missing required field and types in the json schema definition for spec.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -192,7 +192,7 @@
           "$ref": "#/definitions/filter"
         }
       },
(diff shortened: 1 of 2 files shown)


tsurdilo:
- removed filter from required
- used the json schema if/then/else construct to define "nextState" only to be required if "end" is set to false (which is its default value). If "end" is true, nextState is not required.

cathyhongzhang:
I am wondering if we should revert to having an explicit "end state" which seems simpler.

tsurdilo:
@cathyhongzhang: it's pretty simple rule, if a state is not an end state (no end prop defined or end is false) then nextState should be specified. If a state is an end state, it has no next state. This is pretty simple to implement or am i missing something?

PR #110, opened 2019-11-18
fixing function definition

tsurdilo:
currently we have to use function name to define its resource uri. this fixes the issue


cathyhongzhang:
LGTM

PR #111, opened 2019-11-26
Workflow/state explicit error handling

tsurdilo:
Adds explicit runtime exception handling for workflow. Currently the spec does not define any 
runtime exception handling so this is an addition.

### DIFF
--- a/workflow/spec/spec.md
+++ b/workflow/spec/spec.md
@@ -108,6 +108,7 @@ Here we define details of the Serverless Workflow definitions:
 | expressionLanguage |Default expression language to be used throughout the workflow definition | string |no |
 | [triggerDefs](#Trigger-Definition) |Array of workflow triggers | array | no |
 | [states](#State-Definition) | Array of workflow states | array | yes |
+| [onError](#Error-Handling) |Workflow error handling definitions | array | no |
 | extensions | Array of workflow custom extension | array | no |
 
 <details><summary><strong>Click to view JSON Schema</strong></summary>
@@ -179,6 +180,14 @@ Here we define details of the Serverless Workflow definitions:
                 ]
             }
         },
+        "onError": {
+          "type": "array",
+          "description": "Workflow runtime error handling definitions",
+          "items": {
+            "type": "object",
+            "$ref": "#/definitions/error"
+          }
+        },
         "extensions": {
           "type": "array",
           "description": "Workflow Extensions",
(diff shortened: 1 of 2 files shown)


tsurdilo:

tsurdilo:
rebased against master

workflow/spec/spec.md
```diff
@@ -1274,6 +1375,115 @@ Each Filter has three kinds of path filters
...
 
+### Error Handling
+
+Serverless Workflow allows you to explicitly model what should happen in case of runtime errors during workflow execution.
+Explicit error handling is done via the "onError" property which you can place inside a state as well as the main workflow
+definition.
+
+Let's first look at error handling inside states. 
+The states "onError" property is an array containing 
+one ore more error handling definitions. These describe the types of errors which can happen and how serverless workflow
```
mswiderski: tiny typo 'one or more' instead of 'one ore more'

salaboy:
@tsurdilo /LGTM .. we should merge these basic changes

tsurdilo:

manuelstein:
Just curious, but how does the workflow engine obtain the error notification / exception (in case there is an engine in the implementation that orchestrates invocations delivering events to functions using some sort of transport)? The transport layer could obscure what's happening inside the runtime, e.g. a 500 server error with no content. Will the workflow spec go as far as to specify for every runtime and any means of transport as to how runtime-specific stack traces should be communicated?

Another question that came to mind is how timeouts are treated. Are failures to receive any response or transport errors also covered by this or only runtime errors as the description suggests?

salaboy:
@manuelstein errors can be handled as an specific type of event and States can register these Events Listeners for errors. I don't think that we should go deep down to the implementation of transport specific exception handling, this should be more a logical exception handling, that can encapsulate infrastructure errors if needed. 

I would assume that `Runtime Errors` will include time outs and it should follow a similar approach as normal errors, that is, a component that is charge of monitor interaction and emit certain type of events if a timeout is detected, in completing a task.

workflow/spec/spec.md
```diff
@@ -232,6 +241,45 @@ events for same workflow instance, must be specified in that event trigger.
...
+      "type": "string",
+      "description": "Specify result JSON Node of States output as JSONPath. Used to inject error information into state output."
+    },
+    "nextState": {
+      "type": "string",
+      "description": "State to transition to if errors expressed in errorExpression are matched",
+      "minLength": 1
+    }
+  },
+  "required": ["errorExpression", "nextState"]
```
cathyhongzhang: This is a good and must-have addition. I just like to make sure I understand the description correctly. When errorExpression is matched, the error handling function is defined in the nextState. Could you confirm my understanding is correct?

tsurdilo: yeah the logic where and how to continue once that particular runtime exception occurs can de defined in nextState. this can be lets say an operation state calling error handling function, or an entire flowState if needed.

workflow/spec/spec.md
```diff
@@ -232,6 +241,45 @@ events for same workflow instance, must be specified in that event trigger.
...
+{
+  "type": "object",
+  "properties": {
+    "errorExpression": {
+      "type": "string",
+      "description": "Boolean expression to match one or more errors"
+    },
+    "resultPath": {
+      "type": "string",
+      "description": "Specify result JSON Node of States output as JSONPath. Used to inject error information into state output."
```
cathyhongzhang: Could you clarify a little more on what is resultPath? Is it a specification of how to filter the error and then add the filtered result to the state output? If so, it seems we can reuse the filter definition instead of adding a new field. 
I would suggest that you update the description to make it clear. As to reuse of the filter, you can do that later together with update of the filter on the loop state.

tsurdilo:
rebased against master

workflow/spec/spec.md
tsurdilo: @cathyhongzhang good point - updated. take another look pls. 
Also i noticed forgot to do the same for the state looping one. will do a pr asap to fix it there as well.

cathyhongzhang:
LGTM

PR #112, opened 2019-11-26
Fix broken timezone converter link

RainbowMango:
Happen to see the broken links.

Old link with `San Francisco` not work, I guess it's a bug of [timezone converter](http://www.thetimezoneconverter.com/).


duglin:
hmm old one seemed to work for me but ok

Issue #87, opened 2019-10-28
[workflow] Define event-expression format


ruromero:
We need to define how the `event-expression` must be expressed. Trying to keep it language-agnostic.

#### Comparison
* `eq`: equals
* `ne`: not equals
* `empty`: If the value is empty or doesn't exist

#### Logical
* `and`: And operator
* `or`: Or operator
* `not`: Not operator

### Examples

```
name eq 'mytrigger'
type eq 'sometype' or type ne 'uglytype'
not empty(correlation_token)
```

ad-m:
What about [Common Expression Language (CEL)](https://github.com/google/cel-spec)?

tsurdilo:
@ruromero @ad-m the event expression evaluation is an implementation detail.
Implementors can chose for example:
* SPEL 
* JEXL
* MVEL
* Rule engine Evaluation
* etc

We should not force an implementation here at all, but leave it as a simple string value and let implementors decide what expression languages they support or not.

ruromero:
If we want this to be portable between implementations we must specify how to express conditions formally. Otherwise expressions won't work across different implementations.

I agree it could be an existing one. If we decide to support more than one, then there has to be a way to say which one is being used.

e.g.
```
event-expression: "spel: <expression>"
```

I consider this adds complexity and verbosity therefore I prefer to choose one that suits our needs that is not related to any specific programming language. As seen, it doesn't need to be very sophisticated.

The ones proposed by @tsurdilo are Java specific but I am not aware of any popular expression language with implementations of the most common programming languages.

The one mentioned by @ad-m only has cpp and go implementations.

A possible solution would be to define what the expression language used by the implementation must support as in my initial comment and let each spec implementation to pick any expression language implementation that supports this minimum set of operations.

WDYT?

bluemonk3y:
My preference would be to follow SQL as closely as possible. It has universal acceptance and standards.

tsurdilo:
I don't think the specification can say which languages to support or not, but agree that language field if you want can be declared, as is already for example done in bpmn2. So you could have:
```
"expression": {
    "language": "spel",
    "body": "expression-str"
}
```

Note that you should then make the language parameter optional and add it also part of the workflow def parameters, as user might want to use the same language throughout the workflow, and it would be overkill to have to specify it on every expression.

tsurdilo:
fixed via pr https://github.com/cncf/wg-serverless/pull/113. closing.

PR #113, opened 2019-11-28
Adding expression definition

tsurdilo:


tsurdilo:
related to issue https://github.com/cncf/wg-serverless/issues/87

tsurdilo:

salaboy:
@tsurdilo this is a good addition to make sure that the language support language independent expression evaluators.. 
/LGTM

tsurdilo:

manuelstein:
IIUC the eventExpression is meant to be a boolean condition to whether the event should actually cause the trigger to fire. I guess "condition" fits better than "eventExpression" but maybe that's just me.

salaboy:
@manuelstein I tend to agree, so we should have `expressionLanguage` which is kinda correct and `condition` which is where the expressions are defined.  @tsurdilo what do you think about that?

workflow/spec/spec.md
```diff
@@ -375,6 +380,41 @@ should get triggered after this event completes.
...
+
+
+<details><summary><strong>Click to view JSON Schema</strong></summary>
+
+```json
+{
+  "type": "object",
+  "description": "Defines the language and body of expression.",
+  "properties": {
+    "language": {
```
cathyhongzhang: Does this "language" refer to the "expressionLanguage"? If so, it needs to be changed.

workflow/spec/spec.md
```diff
@@ -375,6 +380,41 @@ should get triggered after this event completes.
...
+      "type": "string",
+      "description": "Expression language. For example 'spel', 'jexl', 'cel', etc"
```
cathyhongzhang: Add "json" as an example.

workflow/spec/spec.md
```diff
@@ -375,6 +380,41 @@ should get triggered after this event completes.
...
+    },
+    "body": {
+      "type": "string",
+      "description": "The expression body"
+    }
```
cathyhongzhang: Add "For example, (event1 or event2) and event3"

workflow/spec/spec.md
```diff
@@ -319,7 +324,7 @@ Event state can hold one or more events definitions, so let's define those:
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| eventExpression |Expression used to associate triggerDefs with event state  | string | yes |
+| [eventExpression](#Expression-Definition) |Expression used to associate triggerDefs with event state  | object | yes |
```
cathyhongzhang: Suggest to change to "Expression consisting of Boolean operation of events that will trigger the event state"

workflow/spec/spec.md
tsurdilo: json is not an expression language, so not sure why?

workflow/spec/spec.md
tsurdilo: fixed

workflow/spec/spec.md
tsurdilo: fixed

workflow/spec/spec.md
tsurdilo: fixed

tsurdilo:
@manuelstein updated names according to your suggestions.

workflow/spec/spec.md
tsurdilo: added

PR #114, opened 2019-11-29
Updating functional flow image

tsurdilo:
Small update to image in the Functional Flow section. The existing image was fine, but the connections between states and functions was confusing (it does not show difference between a flow transition and an actual action / function call).  The updated image I think show this clearer. 

Here is old:

Screen Shot 2019-11-29 at 11 06 52 AM

And new:

Screen Shot 2019-11-29 at 11 24 23 AM


salaboy:
@tsurdilo beautiful 
/LGTM

tsurdilo:

manuelstein:
IMHO the function attached to State 1 should _not_ have an arrow to the choice/gateway/decision point/diamond thingy.

AFAICS the old flow could not be represented using the existing event types as State 1 can be either an operation or a choice but IIRC not both.

I think the text in the old one was better readable.

salaboy:
@manuelstein that is true about `State 1` and that should be fixed.
A bigger font for state names and event names would be appreciated .. I agree

cathyhongzhang:
Agree with manuelstein. I think the text in the old one was better readable. Not LGTM.

tsurdilo:
I have updated the image to reflect your comments, here it is 
Screen Shot 2019-12-03 at 12 25 35 PM

Please not that yeah readability might be an issue but that image is clickable where it display the big/real size which is more than readable to users. We have just adjusted the display of the images to be small because of document size. 

Also not that yeah you might like the original image more which is of course fine, but its incorrect to use the same connection types from state to state and to state to function calls..so to me this is more important to fix than some readability problem :)

cathyhongzhang:
LGTM

PR #119, opened 2019-12-06
update state transitions

tsurdilo:
This pr updates state transitions definition. 
Instead of just having "nextState", this adds a "transition" object with two properties:
"nextState" and "condition" (optional)

The first one works as before, however "condition" allows users to restrict/guard transitions by evaluating the states data output.

This is important so that we can prevent harmful transitions from happening.
Also this gives the ability to in future secure transitions etc.

An example is given in the spec description

This also fixes some existing examples json.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -114,7 +114,7 @@
       "type": "object",
       "properties": {
         "condition": {
-          "description": "Boolean expression which consists of one or more Event operands and the Boolean operators",
+          "description": "Boolean expression evaluated against state's data output. Must evaluate to true for the transition to be valid.",
           "$ref": "#/definitions/condition"
         },
         "nextState": {
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -110,6 +110,23 @@
     "states"
   ],
   "definitions": {
+    "transition": {
+      "type": "object",
+      "properties": {
+        "condition": {
+          "description": "Boolean expression which consists of one or more Event operands and the Boolean operators",
+          "$ref": "#/definitions/condition"
+        },
+        "nextState": {
+          "type": "string",
+          "description": "State to transition to next",
+          "minLength": 1
+        }
+      },
+      "required": [
+        "nextState"
+      ]
+    },
     "error": {
       "type": "object",
       "properties": {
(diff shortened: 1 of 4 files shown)


tsurdilo:
rebased vs master

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -110,6 +110,23 @@
     "states"
   ],
   "definitions": {
+    "transition": {
+      "type": "object",
+      "properties": {
+        "condition": {
+          "description": "Boolean expression which consists of one or more Event operands and the Boolean operators",
+          "$ref": "#/definitions/condition"
```
cathyhongzhang: The spec states that "Conditions can define an expression against the states output data". The description here does not match it. It is not clear whether the condition is based on events or on the output data of the state.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -110,6 +110,23 @@
     "states"
   ],
   "definitions": {
+    "transition": {
```
cathyhongzhang: I feel "next state" is more self explanatory. I would suggest that we replace "transition" with "next state" which has two properties: "condition" and "state".

workflow/spec/spec.md
```diff
@@ -256,7 +256,7 @@ see the [Error Handling section](#Error-Handling).
 | --- | --- | --- | --- |
 | errorExpression | Boolean expression to match one or more errors | string |yes |
 | [filter](#Filter-Definition) |Error data filter | object | yes |
-| nextState |State to transition to if errors expressed in errorExpression are matched | string | yes |
+| [transition](#Transitions) |Transition definition when errors expressed in errorExpression are matched | string | yes |
```
cathyhongzhang: It is really "the next state to transition to (if there is a condition, then condition needs to be satisfied)", not "Transition definition(what is added in the schema.json is the transition definition)". I think it is better to call it "nextState"

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: i think "transition" is much clearer sorry :) as the workflow makes a "transition" from A to B. It may be that those are not always "states" in the future.

workflow/spec/spec.md
tsurdilo: transition is less restrictive than "nextState". we need a much broader and more powerful construct and "transition" will provide that

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: updated. thanks.

workflow/spec/spec.md
cathyhongzhang: Ok, Then can we say "the next transition of the workflow when ..." or something like that instead of "transition definition"

tsurdilo: good point. will update all docs to say that

tsurdilo: updated everything to state what you mentioned. thanks!

tsurdilo:
rebased

cathyhongzhang:
LGTM

PR #124, opened 2019-12-17
Workflow Data Section

tsurdilo:
This pr adds the "Workflow Data" section which replaces the 
"Information Passing" and "Filter Mechanism" sections (adds to them, does not remove anything).
It also replaces and unifies all images and updates the table of contents.

Reason: the current specification poorly describes workflow data and its manipulation during workflow execution. This updates that and makes things much clearer for readers.

You can see what the changes look like as markup (much easier to review) here: 
https://github.com/tsurdilo/wg-serverless/blob/initcontext/workflow/spec/spec.md#workflow-data


salaboy:
@tsurdilo I do understand where is this coming from, but I am starting to wonder if we really need to cover data flow handling in here. After checking with some people, DataFlow in Serverless is more associated to streams of data flowing through processing functions, and in this case we are talking about JSON payloads that contains maps of simple data.  I wonder if we should consider treating Data Flow as an addition to the base spec and not a mandatory part of it. I hope it makes sense. I am happy to do a quick call to catch up on this one..

tsurdilo:
@salaboy i think we have to cover basics and definitely work on the streaming part as an addition + revisit the basics when we have that in place. 
This change keeps pretty much everything in-tact as it was before (the basics) and adds workflow data input/output as we have to have that to cover usability cases with some of the existing workflow solutions out there.

I am all for you leading the streaming data part of this spec and am looking forward to learn that part, but still think we have to have this basic in spec as workflow has to be usable with/without streams (imo). 

wdyt?

mswiderski:
I actually like this rewrite as it makes things more explicit but at the same time I don't see any change in the spec itself. So +1 from my end as it helps to better understand what is expected and available from data point of view in the workflows and not enforce any of them at all.

tsurdilo:
rebased

salaboy:
@tsurdilo @mswiderski I am happy with these changes /lgtm , but I am still thinking that we need to separate these into a separate/sub spec. We all know how painful data handle is and even though I know that this is needed in real use cases, having this as optional might lead to a conceptual simplification for the initial iteration of the spec.

PR #128, opened 2019-12-23
New state - Relay state

tsurdilo:
We have a need to create a no-op state which can be used to set up static data which users can test the behavior of their workflow logic with. 

This state is similar to the Script task in bpmn or the Pass state in aws.
It is very useful for debugging workflow logic/execution and is really nice to have for examples as it can be much clearer to users than assuming some dynamic data returned from actions/functions etc.

This state is also useful not only for debugging but for production apps where we want to only transform workflow data and not do any work.

The pr includes a detailed state description, its json schema definition, and couple examples of its use.


workflow/spec/spec.md
```diff
@@ -1365,6 +1368,164 @@ Each sub-workflow receives a copy of the SubFlow state's input data.
...
 If this property is set to false, data access to parent's workflow should not be allowed.
 
+### Transform State
+
+| Parameter | Description | Type | Required |
+| --- | --- | --- | --- |
+| id | Unique state id | string | no |
+| name |State name | string | yes | 
+| type |State type | string | yes | 
+| inject |JSON object which can be set as states data input and can be manipulated via filter | object | no |
```
cathyhongzhang: typo: state data input->state's data input.

workflow/spec/spec.md
```diff
@@ -1365,6 +1368,164 @@ Each sub-workflow receives a copy of the SubFlow state's input data.
 If waitForCompletion property is set to true, sub-workflows have the ability to edit the parent's workflow data.
 If this property is set to false, data access to parent's workflow should not be allowed.
 
+### Transform State
+
```
cathyhongzhang: how about call it "relay" state?

workflow/spec/spec.md
```diff
@@ -321,7 +322,9 @@ States define building blocks of the Serverless Workflow. The specification defi
 - **[Parallel State](#Parallel-State)**: Allows a number of states to execute in
     parallel.
     
-- **[SubFlow State](#SubFlow-State)**: Allows execution of a sub-workflow.   
+- **[SubFlow State](#SubFlow-State)**: Allows execution of a sub-workflow. 
+  
+- **[Transform State](#Transform-State)**: Used to transform state data input and output without performing any work.
```
cathyhongzhang: If I understand this state correctly, suggest to change to: 
Used to relay state's data input to output without executing any serverless function although the input data may run through a filter

workflow/spec/spec.md
```diff
@@ -1365,6 +1368,164 @@ Each sub-workflow receives a copy of the SubFlow state's input data.
...
+    },
+    "else": {
+      "required": ["name", "type", "transition"]
+    }
+}
+```
+
+</details>
+
+Transform state is a no-op state which can be used to statically set up and transform data input and output.
```
cathyhongzhang: Transform state can be used to statically set up and relay data input to output.

workflow/spec/spec.md

workflow/spec/spec.md
tsurdilo: ok . will change.

workflow/spec/spec.md

workflow/spec/spec.md
tsurdilo: updated. changed a little from suggestion

tsurdilo:
@cathyhongzhang thanks for review - updated per your suggestions.

tsurdilo:
added relay state example as well

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -819,6 +827,53 @@
...
+          },
+          "filter": {
+            "$ref": "#/definitions/filter"
+          },
+          "transition": {
+            "description": "Next transition of the workflow after subflow has completed",
+            "$ref": "#/definitions/transition"
+          }
+        },
+        "if": {
```
cathyhongzhang: Could you clarify why and how if, then, else will be used? Adding more description will help people understand better why they are needed.

cathyhongzhang: never mind. I got it.

tsurdilo: ok :) so is this good to merge now or needs still some changes?

cathyhongzhang:
LGTM

PR #132, opened 2019-12-30
Action definitions - externalizing actions

tsurdilo:
Actions are currently used within both event and operation states. As they are defined, same actions cannot be reused between these two states (have to be defined multiple times).

This pr externalizes actions into "actionDefs" section of the workflow (similar to triggerDefs).

Event and Operation states can then instead of defining each individual action reference a set of defined actions by their action name.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -255,10 +255,6 @@
           "type": "string",
           "description": "Action unique name"
         },
-        "group": {
-          "type": "string",
-          "description": "Action group name"
-        },
         "function": {
           "description": "Specifies the function that must be invoked",
           "$ref": "#/definitions/function"
@@ -277,8 +273,7 @@
       },
       "required": [
         "name",
-        "function",
-        "group"
+        "function"
       ]
     },
     "retry": {
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -53,6 +53,14 @@
         "$ref": "#/definitions/triggerevent"
       }
     },
+    "actionDefs": {
+      "type": "array",
+      "description": "Action Definitions",
+      "items": {
+        "type": "object",
+        "$ref": "#/definitions/action"
+      }
+    },
     "states": {
       "type": "array",
       "description": "State Definitions",
@@ -219,10 +227,10 @@
         },
         "actions": {
           "type": "array",
-          "description": "Action Definitions",
+          "description": "String array containing the defined action group names to be executed. Groups are executed in order defined.",
+          "default": [""],
           "items": {
-            "type": "object",
-            "$ref": "#/definitions/action"
+            "type": "string"
           }
         },
         "transition": {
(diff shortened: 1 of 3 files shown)


workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -215,10 +223,10 @@
         },
         "actions": {
           "type": "array",
-          "description": "Action Definitions",
+          "description": "String array containing the defined action group names to be executed. Groups are executed in order defined.",
```
cathyhongzhang: It is not clear to me whether the "actions" is an array of action groups or an array of actions. The descriptions says it is an array of action groups while the definition shows it is an array of actions if I read it correctly. I understand your point of reuse which I support. Existing spec already supports reuse of functions. I am not sure if the benefit of supporting reuse of groups justifies the complicated layers of definition. When I go through the syntax, I am kind of lost in the layer relationship and syntax matching. For people who are new to this spec, they can be lost and be confused too. I would suggest that for now let's keep it as is. We can add the group later if we see that many use cases need it. Wdyt?

tsurdilo: i think i understand your point -- still think externalizing actions is very beneficial, if groups are too abstract at this point, I can remove and users would have to list actions to be ran by name . would that be ok with you ?

cathyhongzhang: Do you mean that you will not add the group and close this pr? If so, yes, I think users can reuse the function.

tsurdilo: No i still think that reusing actions is a big benefit, as currently both event states and operation states have them. I meant that the action grouping to be removed as that part you say is confusing, and in "actions" property users would call actions by name instead of group names. I'll make this update shortly and it should be more clear then

tsurdilo: and btw. actions cannot be reused currently, not sure how you think thats possible. The trigger definitions are via the event expression, but thats just for event states atm.

tsurdilo: we could go even further and add functionDefs (or just "functions" if we want to add reusability of functions as well. That would be a good idea

tsurdilo:
@cathyhongzhang updated to remove the actionDefs "group" property. Now event and operation states have to define the action name(s) to be used and not groups. Hope this is much cleaner now.

tsurdilo:
rebased

workflow/spec/schema/serverless-workflow-schema-01.json
cathyhongzhang: Have you updated with new change? I think function can be reused, but maybe not action. Action has "timeout", "retry" and "filter", which may need to be changed when a function is reused in another state or another workflow. wdyt?

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -243,6 +251,14 @@
     "action": {
       "type": "object",
       "properties": {
+        "name": {
+          "type": "string",
+          "description": "Action unique name"
+        },
+        "group": {
```
cathyhongzhang: have you updated? I still see the group. Maybe I misunderstand something here? Thanks.

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: sorry there was some merge issue. updated now . good point about reusing functions. thats better idea. I will close this and add reuse of functions instead. thanks

PR #135, opened 2020-01-05
Updating transition definition - adding 'end' property

tsurdilo:
It is actually really useful to be able to end workflow execution inside the transition definition. 
The pr includes two examples that you can look at to see why.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -121,11 +121,24 @@
           "type": "string",
           "description": "State to transition to next",
           "minLength": 1
+        },
+        "end": {
+          "type": "boolean",
+          "default": false,
+          "description": "Instead of transitioning to nextState, end workflow execution instead"
         }
       },
-      "required": [
-        "nextState"
-      ]
+      "if": {
+        "properties": {
+          "end": { "const": true }
+        }
+      },
+      "then": {
+        "required": ["end"]
+      },
+      "else": {
+        "required": ["nextState"]
+      }
     },
     "error": {
       "type": "object",
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -125,7 +125,7 @@
         "end": {
           "type": "boolean",
           "default": false,
-          "description": "Instead of transitioning to nextState, end workflow execution instead"
+          "description": "Instead of transitioning to nextState, end workflow execution"
         }
       },
       "if": {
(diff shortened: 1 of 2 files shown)


workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -121,11 +121,24 @@
           "type": "string",
           "description": "State to transition to next",
           "minLength": 1
+        },
+        "end": {
+          "type": "boolean",
+          "default": false,
+          "description": "Instead of transitioning to nextState, end workflow execution"
         }
```
cathyhongzhang: I am not sure if adding another end in the transition is good. Operation state has "end" param and "transition" param. If "transition" also has an "end" param, it may bring more human errors since people might specify conflicting "end" in the spec by mistake. People prefer a simple and easy to use spec. wdyt?

tsurdilo: maybe "end" is not the best name for it? "complete" maybe? Please look at examples to see why. this makes things simpler -- it is imo alot more confusing if users have to add bogus end states

cathyhongzhang: Maybe I misunderstand you. But whatever we name it, this opens up for potential human errors of specifying conflicting "end" for operation state, and maybe for other scenarios that use the transition. From my past experience, once a spec provides two options, people may have questions on where is the correct place they should specify the "end".  I feel a clear end state is less confusing even its functionality is just a bogus end.

tsurdilo: @cathyhongzhang ok.I think that's fair. wanted to ask before closing this pr then - can you think of some other more intuitive way inside choices of switch state state for example to say:

"for this choice just end workflow execution"?

cathyhongzhang: Not sure what you mean. Now that we have replaced next state with "transition", if a choice needs to end a workflow, the choice should transit to an state that is an end state while other choices can transit to some intermediate states before transiting to an end state.  BTW, can a  switch state itself be an end state? Probably not since the definition of a switch state is to transit to another state. If so, we need to remove the "end" field. Wdyt? I can create a PR to remove it.

tsurdilo: yes, choice state can be an end state:
if no choices match, "end": true will terminate the workflow
if no end is specified, users can set the "default" property  to transition to a new state...
looking at it tho "default" should be a transition not a string..i missed that when doing the transition pr..will update asap

cathyhongzhang: Ok, Thanks!

PR #139, opened 2020-01-07
Reusable functions - referenced in actions

tsurdilo:
Functions are now defined similar to triggers and can be referenced within actions of event and operation states. 
Makes functions reusable between defined actions.
Functions are referenced via "functionref" property in actions. Function properties are defined inside functionref of actions.

Also this pr:
* changes "triggerDefs" to "triggers"
* renames retry definition "match" to "condition" so its a proper condition and not a string


cathyhongzhang:
LGTM

Issue #141, opened 2020-01-08
restructure the event state and operation state


cathyhongzhang:
Here is my thought to address the confusion on the event state: 
We can do the following to make it more clear and more consistent with the definition of other states, especially consistency with the Operation state
1. rename the "events" field to "eventAction" since the definition of "events" includes not only event triggers, but also actions and action mode association with the event triggers, and the transition to next state.  Suggestion for a better name is welcomed. 
2. Add "operationAction" to operation state and move "actionMode" and "actions" inside "operationAction". This will make the first level fields of all the states be more consistent.
3. Modify the "condition" definition inside existing "events" as follows:
   "Condition consisting of Boolean operation of events that will trigger the actions"
   I guess existing description that says "Condition consisting of Boolean operation of events that will trigger the event state" could cause confusion. If the event state is not the start state, the previous state transition triggers transition to the event state, not the events. 

wdyt?

tsurdilo:
@cathyhongzhang here are changes i did based on your info here: https://github.com/cncf/wg-serverless/pull/142

1. called the parameter "onReceive"
2. this just brings another layer of json definition which imo is not needed for operation state. didnt do this
3. updated to "Condition matching a received event with one or more defined triggers"

cathyhongzhang:
The reason of bringing another layer of json is to make the first level of state definition to be consistent among all the states. If you look at the first level definition of Event state, switch state etc. all have two levels of definition, except operation state. But I do not have a strong opinion on this.

tsurdilo:
Can be closed as https://github.com/cncf/wg-serverless/pull/142 is merged

PR #142, opened 2020-01-08
Updating event and operation states - issue #141

tsurdilo:


tsurdilo:
Changes based on issue https://github.com/cncf/wg-serverless/issues/141.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -206,35 +206,36 @@
         "inputPath"
       ]
     },
-    "event": {
+    "onreceive": {
       "type": "object",
       "properties": {
         "condition": {
-          "description": "Boolean expression which consists of one or more Event operands and the Boolean operators",
+          "description": "Condition matching a received event with one or more defined triggers",
```
cathyhongzhang: I think existing description gives more info about what condition means. The definition of condition is a Boolean expression of trigger events.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -206,35 +206,36 @@
         "inputPath"
       ]
     },
-    "event": {
+    "onreceive": {
```
cathyhongzhang: I feel onreveive is a more confusing name. onreceive what?

tsurdilo: event(s) that match defined conditions

cathyhongzhang: I understand, but "onreceive" seems to convey less info and more ambiguous. I don't feel this is a better change. Let's think more and try to come up with a better naming.

tsurdilo: Any ideas? I like the current "onReceive". To me at least it's not confusing. 
@salaboy any suggestions?

tsurdilo:
rebased against master

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: updated

tsurdilo:
any ideas ..otherwise i'd like to move this forward please and listen to suggestions about different names when they come..we can update then.

cathyhongzhang:
I think "eventAction" is a better name, which matches the description better "Define what events to act upon and actions to be performed". If you do not agree, my suggestion is to not change it (keep as is).

tsurdilo:
@cathyhongzhang its an array of actions so "eventActions" ok for you ?

cathyhongzhang:
Yes, eventActions is good for me.

cathyhongzhang:
LGTM

PR #143, opened 2020-01-09
New State - ForEach

tsurdilo:
As discussed we wanted to remove looping from being part of each individual state into its own state construct. 

This PR adds the "ForEach" state which is used to iterate over a JSON array in its data input. 
The documentation is spec.md has its schema definition, parameter description as well as an example. 

Also updated spec-examples.md updating the looping example to use ForEach state instead of "loop" parameter.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -919,9 +919,15 @@
             "type": "string",
             "description": "JSONPath expression specifying where in the states data output to place the final data output of each iteration of the executed states"
           },
-          "completionCondition": {
-            "$ref": "#/definitions/condition",
-            "description": "Boolean condition (evaluated after each iteration) evaluated against the inputParameter. If this condition is true iterations stops even if not all elements of inputCollection are processesd"
+          "inputParameter": {
+            "type": "object",
+            "description": "JSONPath expression specifying an JSON object field of the states data input. For each parallel iteration, this field will get populated with a unique element of the inputCollection array"
+          },
+          "max": {
+            "type": "integer",
+            "default": "0",
+            "minimum": 0,
+            "description": "Specifies how upper bound on how many iterations may run in parallel"
           },
           "timeDelay": {
             "type": "string",
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -94,6 +94,10 @@
           {
             "title": "Relay State",
             "$ref": "#/definitions/relaystate"
+          },
+          {
+            "title": "ForEach State",
+            "$ref": "#/definitions/foreachstate"
           }
         ]
       }
@@ -379,6 +383,10 @@
                 {
                   "title": "Relay State",
                   "$ref": "#/definitions/relaystate"
+                },
+                {
+                  "title": "ForEach State",
+                  "$ref": "#/definitions/foreachstate"
                 }
               ]
             }
@@ -427,9 +435,6 @@
           "timeDelay": {
             "type": "string",
             "description": "Amount of time (ISO 8601 format) to delay"
-          },
-          "loop": {
-            "$ref": "#/definitions/loop"
           },
           "onError": {
             "type": "array",
@@ -504,9 +509,6 @@
               "$ref": "#/definitions/eventactions"
             }
           },
-          "loop": {
-            "$ref": "#/definitions/loop"
-          },
           "onError": {
             "type": "array",
             "description": "States runtime error handling definitions",
@@ -566,9 +568,6 @@
               "type": "object",
               "$ref": "#/definitions/action"
             }
-          },
-          "loop": {
-            "$ref": "#/definitions/loop"
           },
           "onError": {
             "type": "array",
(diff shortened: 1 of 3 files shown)


workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -896,6 +886,114 @@
...
+                {
+                  "title": "SubFlow State",
+                  "$ref": "#/definitions/subflowstate"
+                },
+                {
+                  "title": "Relay State",
+                  "$ref": "#/definitions/relaystate"
+                },
+                {
+                  "title": "ForEach State",
```
cathyhongzhang: Do we really need embedded ForEach state? I feel we are making it too complicated for the user to understand and use.

workflow/spec/spec.md
```diff
@@ -199,11 +199,38 @@ Here we define details of the Serverless Workflow definitions:
...
+                      "title": "SubFlow State",
+                      "$ref": "#/definitions/subflowstate"
+                    },
+                    {
+                      "title": "Relay State",
+                      "$ref": "#/definitions/relaystate"
+                    },
+                    {
+                      "title": "ForEach State",
+                      "$ref": "#/definitions/foreachstate"
```
cathyhongzhang: I feel "loop" state is a better name.

workflow/spec/spec.md
```diff
@@ -379,7 +406,10 @@ States define building blocks of the Serverless Workflow. The specification defi
 - **[SubFlow State](#SubFlow-State)**: Allows execution of a sub-workflow. 
   
 - **[Relay State](#Relay-State)**: Used to relay state's data input to output without executing any actions. State's data input can be filtered. 
-    
+
+- **[ForEach State](#ForEach-State)**: Allows a set of defined states to be executed for each element of a data input array.
+
```
cathyhongzhang: I think we need more clarification on this state. When I read it, I am not sure if this means that the data input array has multiple elements and each element will trigger transition to a state, .i.e. it can transit to multiple states in parallel. Is this correct? Could the state being transited to different? Is it a input array or just a filtered part of the input data?

workflow/spec/spec.md
tsurdilo: I believe foreach is the most widely used. It also imo better suited for its functionality

workflow/spec/spec.md
tsurdilo: There is a detailed description and example included later on. This is just a short description. 

* Its explained that iterations happen over an array element of the data input
* For each iteration the defined states are executed. There is a startsAt property which determines which is the start state.
* This pr includes description of transitions outside are not allowed and outside transitions to states inside foreach are not allowed
* it is already described and shows in example that the array is an element of the data input not the data input itself. You could make it the data input itself if you set the inputCollection to "$."

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: Is our goal really to dumb this spec down for people or allow them to orchestrate their serverless application how they want? :) 

I can remove it np.....

tsurdilo: there could be valid business reasons for example:
"Do X for each completed order and do Y for each item in that order"

tsurdilo: this actually gave me a nice idea for a new example :)

tsurdilo:
rebased against master

tsurdilo:
can we finish the review please? PRs are sitting in queue again

tsurdilo:
rebased

tsurdilo:
@cathyhongzhang i made some updates the the state documentation to make things clearer for readers, let me know what you think.

workflow/spec/spec.md
```diff
@@ -1602,131 +1632,259 @@ $.people[?(@.age >= 40)]
...
 }
 ```
 
 </details>
 
-Filters are used for managing workflow data flow. 
-This is described in detail in the [Information Passing](#Information-Passing) section.
-
-### Loop Definition
+The ForEach state can be used to execute a defined set of states for each element of an array (defined in the states data input).
```
wanghq: Some questions:

1. Seems the foreach state executes the defined steps for each entry **sequentially**. Is there any plan to support parallel foreach? Is that going to be another state, or use the same state with some fields indicating that?
2. Seems for the given example, there is no way to use the result of previous iteration in the current iteration (e.g. sum the expressions), while foreach in any common programming lang allows that. But for a parallel foreach state, each iteration should be independent (because of the data race problem).

So I feel the foreach defined here is more like parallel iterations and there is no need to emphasize the sequential nature. If sequential iterations are needed, having an attribute like concurrency to control the behavior would be good.

tsurdilo:
@wanghq thanks for comments

updated foreach state to define parallel executions (as it should have from beginning as you mentioned). So regarding your comments:

1. Resolved - pls take another look
2. For parallel execution does not apply as mentioned

workflow/spec/spec.md
salaboy: +1 with foreach

workflow/spec/spec.md
salaboy: +1

tsurdilo:
rebased

PR #146, opened 2020-01-21
Fixing workflow schema

tsurdilo:
Fixes schema syntax error and reformats it


tsurdilo:
This is a big error that would break community tooling. So I will go ahead and push it through

Issue #130, opened 2019-12-30
[workflow] Proposal: Serverless Workflow WG Scope


salaboy:
As presented in San Diego, the scope of the group has been the definition of a vendor-neutral Workflow Language for Serverless Applications. 

As stated in the main README file of the group: 

> “The goal of the Serverless Workflow sub-group is to come up with a standard way for users to specify their serverless application workflow, as well as help, facilitate portability of serverless applications across different vendor platforms. 
> Serverless Workflow is a vendor-neutral and portable specification that meets these goals.” 

While this is a big effort in itself, having just a language will not promote cloud providers to implement it, we need to define how the language and the concepts defined in the group will be mapped to actual implementations in the existing CNCF ecosystem. 

As stated in [workflow Proposal: Confusing Terms and Scope of Workflow WG · Issue #127 · cncf/wg-serverless · GitHub](https://github.com/cncf/wg-serverless/issues/127), the lack of clear scope and how the current language will be handled by different vendors is causing confusion, hence this proposal. 

# Proposal
The scope of the group should be changed to cover two main angles:
- Workflow Language (well-scoped for the first iteration)
- How cloud providers can integrate with Workflows, meaning which standard APIs/Contracts should Cloud Providers implement in order to integrate with workflows. 

By covering these two points we will promote (as a group) the use of the vendor-neutral language as well as how cloud providers can understand these workflows. 

In a Kubernetes world, the integration with the ecosystem can be done by defining a new concept (CRD) “Workflow” that Kubernetes can understand and then manage and monitor the lifecycle of each Workflow. 

By providing this new concept, each cloud provider can understand that there is a new Workflow available and then do whatever is necessary to deploy, execute and monitor these Workflows. 
The “Workflow” concept itself can include a reference to the workflow definition and metadata that can be used as an envelope for the definition itself. 

The “Workflow” concept becomes part of a Standard API which is vendor-neutral as well as the Workflow Definition Language. 

An initial proposal is provided in this PR: 
https://github.com/cncf/wg-serverless/pull/131 

Feedback is more than welcome.

tsurdilo:
@salaboy can this issue be closed as https://github.com/cncf/wg-serverless/pull/131 is merged ?

salaboy:
@tsurdilo I would like to include a description in the group motivation as well.. is it ok if I send a PR there?

salaboy:
[...]

PR #147, opened 2020-01-23
Updating description with larger, but well-scoped goals

salaboy:
Goals described here: 
https://github.com/cncf/wg-serverless/issues/130

### DIFF
--- a/workflow/spec/readme.md
+++ b/workflow/spec/readme.md
@@ -10,8 +10,11 @@ the execution order (steps) for countless functions triggered by as many events.
 Workflows have become key components of serverless applications as they excel at orchestration and coordination
 of their functional flow. 
 
-The goal of the Serverless Workflow sub-group is to come up with a standard way for users to specify their serverless application workflow, as well as help facilitate 
-portability of serverless applications across different vendor platforms.
+The goal of the Serverless Workflow sub-group is two-fold:
+-  Come up with a standard language for users to specify their serverless application workflow, as well as 
+- Provide a Kubernetes Standard API to integrate Workflows into the ecosystem
+
+This will facilitate the portability of serverless applications across different vendor platforms.
 
 Serverless Workflow is a vendor-neutral and portable specification which meets these goals.
 
--- a/workflow/spec/readme.md
+++ b/workflow/spec/readme.md
@@ -11,8 +11,8 @@ Workflows have become key components of serverless applications as they excel at
 of their functional flow. 
 
 The goal of the Serverless Workflow sub-group is two-fold:
--  Come up with a standard language for users to specify their serverless application workflow, as well as 
-- Provide a Kubernetes Standard API to integrate Workflows into the ecosystem
+- Come up with a standard language for users to specify their serverless application workflow, as well as 
+- Provide a Cloud Provider agnostic API to integrate Workflows into the ecosystem
 
 This will facilitate the portability of serverless applications across different vendor platforms.
 
--- a/workflow/spec/readme.md
+++ b/workflow/spec/readme.md
@@ -12,7 +12,7 @@ of their functional flow.
 
 The goal of the Serverless Workflow sub-group is two-fold:
 - Come up with a standard language for users to specify their serverless application workflow, as well as 
-- Provide a Cloud Provider agnostic API to integrate Workflows into the ecosystem
+- Provide a Cloud Provider agnostic API to integrate Workflows into the ecosystem.
 
 This will facilitate the portability of serverless applications across different vendor platforms.
 


workflow/spec/readme.md
```diff
@@ -10,8 +10,11 @@ the execution order (steps) for countless functions triggered by as many events.
 Workflows have become key components of serverless applications as they excel at orchestration and coordination
 of their functional flow. 
 
-The goal of the Serverless Workflow sub-group is to come up with a standard way for users to specify their serverless application workflow, as well as help facilitate 
-portability of serverless applications across different vendor platforms.
+The goal of the Serverless Workflow sub-group is two-fold:
+-  Come up with a standard language for users to specify their serverless application workflow, as well as 
+- Provide a Kubernetes Standard API to integrate Workflows into the ecosystem
```
tsurdilo: I would strongly prefer if we keep integration/implementation/technology related things out of the readme and spec documents. I think a suited place for this are the infrastructure/community/reference docs. If you can please move them there.

workflow/spec/readme.md
```diff
@@ -10,8 +10,11 @@ the execution order (steps) for countless functions triggered by as many events.
 Workflows have become key components of serverless applications as they excel at orchestration and coordination
 of their functional flow. 
 
-The goal of the Serverless Workflow sub-group is to come up with a standard way for users to specify their serverless application workflow, as well as help facilitate 
-portability of serverless applications across different vendor platforms.
+The goal of the Serverless Workflow sub-group is two-fold:
```
tsurdilo: The goals are clearly described in the specification document. I think if we put them here in readme which people are going to read first before going to the spec document, they will get out of sync. 
Also "serverless application workflow" does not mention "orchestration", all things that we have already in the spec document, So overall I would prefer if we don't include this here at all.

workflow/spec/readme.md
tsurdilo: We do not want people to think that there is any intention to make this specification in any way shape or form tied to a particular container or orchestration systems. We can add configurations for the major ones, API/SPI for different languages etc, but the specification is **completely** neutral as far as impl / deployment is concerned. Please lets keep it that way.

cathyhongzhang: Agree with Tihomir's points

mswiderski: +1 to this point to keep it agnostic of both implementation and target runtime environment

salaboy: @cathyhongzhang @mswiderski, thanks a lot for your inputs here. 
I've discussed it with @tsurdilo and the issue seems to be the word Kubernetes, so I will replace that to make it more agnostic.

workflow/spec/readme.md
salaboy: @tsurdilo I do still disagree on this. One is the scope of the group and the other is the goal of the language spec. Can we make that difference? should we do that in a premier document as @duglin suggested?

salaboy: @tsurdilo my proposal will be to have a link into the first point to the language spec and in the second point a link to cloud provider interoperability

workflow/spec/readme.md
salaboy: replaced

tsurdilo:
@salaboy thanks for making the updates. Just have a question regarding:
** Provide a Cloud Provider agnostic API to integrate Workflows into the ecosystem **
Do you here describe different specification-provided SDK implementations, such as what CloudEvents has, for example:
https://github.com/cloudevents/sdk-go
https://github.com/cloudevents/sdk-csharp
https://github.com/cloudevents/sdk-java
etc?

If so I think this is a much much bigger task to do compared to CloudEvents:
1) workflow spec is much bigger in scope
2) many different implementations (runtime/api/spi) already exist and are applied by existing serverless workflow solutions.

We can of course like CloudEvents provide different SDKs for different languages, but we only can imo confidently require compliance to spec on the language definition level. 
We cannot possibly ask Amazon or Microsoft to have to change their existing apis/spis to conform to the specification (or anyone else) and stating that as a main goal of this specification I think does not promote adoption of it at all. WDYT?

salaboy:
@tsurdilo I think that your comment is exactly on point about why this is getting confusing and the main reason that I used the term "Kubernetes" before. In the cloud space, Kubernetes is the Abstraction between Cloud Providers, which means that if you build something for Kubernetes, as the CRD that I've provided, you are pushing for something that is Cloud Vendor agnostic and portable. 

And to your question, Amazon and Microsoft both changed their Cloud APIs to Kubernetes, so at that level yes. And that is the level that I am aiming with the infrastructure angle here in the group. Remember, this is not about the language, it is about how cloud providers understand the Workflow concept and then when there is a definition, how do they treat them. To your point, I am pretty sure that we will not be able to push any of those companies to adopt a new language for the function orchestration, but at least we can make the interaction portable and use the language for portability. 

Does this make sense? 

Cheers

salaboy:
Notice that the integration with the infra is completely language-agnostic, it doesn't push for any language implementation at all.

tsurdilo:
@salaboy ok. thanks for explaining!!

LGTM

workflow/spec/readme.md
```diff
@@ -10,8 +10,11 @@ the execution order (steps) for countless functions triggered by as many events.
...
+- Come up with a standard language for users to specify their serverless application workflow, as well as 
+- Provide a Cloud Provider agnostic API to integrate Workflows into the ecosystem
```
tsurdilo: needs period at end?
ecosystem or ecosystems?

salaboy: +1 period in the end. 
I would say that Ecosystem is ok in singluar.. as the Cloud Native Ecosystem

salaboy:
done ! :)

workflow/spec/readme.md
```diff
@@ -10,8 +10,11 @@ the execution order (steps) for countless functions triggered by as many events.
 Workflows have become key components of serverless applications as they excel at orchestration and coordination
 of their functional flow. 
 
-The goal of the Serverless Workflow sub-group is to come up with a standard way for users to specify their serverless application workflow, as well as help facilitate 
-portability of serverless applications across different vendor platforms.
+The goal of the Serverless Workflow sub-group is two-fold:
+- Come up with a standard language for users to specify their serverless application workflow, as well as 
+- Provide a Cloud Provider agnostic API to integrate Workflows into the ecosystem.
+
```
cathyhongzhang: @salaboy I am not sure what you mean by "API to integrate workflows into the ecosystem". Here we define the language primitives for the user to specify their serverless application workflow. Do you mean you will create a set of APIs to map the language primitives into various existing workflow model specifications or do you mean somehting else? What exactly does the ecosystem represent here? Could you clarify it?

salaboy: @cathyhongzhang yeah.. totally, what I've mentioned before here in the thread, it is the CNCF ecosystem, and cloud providers. One thing is the language and then another is how each implementation will run these workflows. Because we are not going that deep into implementation details, having an API that different vendors can pick up is really important to make the language approachable.

salaboy:
@tsurdilo I think that we can close this one then.

salaboy:
👍

cathyhongzhang:
@salaboy Thanks for the clarification.

PR #148, opened 2020-01-23
Removing execStatus property

tsurdilo:
Removing execStatus property as:
* we do not describe currently what it does
* we do no explain how it is set during workflow execution
* we added workflow data input / output which can contain this information

I will revisit this soon and propose an alternative way for it then we can discuss, but for now this property is useless


cathyhongzhang:
LGTM

PR #149, opened 2020-01-23
Event state updates + Event state example

tsurdilo:
This pr updates the event state. The main reason is that we need to clearly define that workflows can both **consume** and **produce** events (producing will be introduced in next pr...)

Event states consume events and perform actions.

Pr includes an event state example as well.


workflow/spec/spec-examples.md
```diff
@@ -214,6 +215,161 @@ states:
 <img src="media/greetingexample.png" with="400px" height="400px" alt="Greeting Example"/>
 </p>
 
+### Event Based Greeting Example
+
+#### Description
+
+This example shows a single Event state with one action that calls the "greeting" function. 
+The event state consumes cloud events of type "greetingEventType". When an even with this type
```
wanghq: Could you explain how this works? The below sample event doesn't contain any **execution** information. If there are many executions of the same workflow, which execution's state is going to consume the event?

tsurdilo: This would depend on the actual implementation which this specification does not go into at all. 

I would assume that of you have 10 instances of this workflow, all 10 that define to consume the event would do so. 

If you create a reference implementation where this should not be the default behavior I would assume you would put that into the reference impl documentation.

Hope this helps.

tsurdilo: Conformance with this specification is on the workflow definition level atm.
I think this is something a lot of people miss and maybe we should document better.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -213,7 +213,7 @@
     "eventactions": {
       "type": "object",
       "properties": {
-        "condition": {
+        "eventCondition": {
           "description": "Boolean expression which consists of one or more Event operands and the Boolean operators",
```
cathyhongzhang: When I see "eventCondition", my first thought is that it is about the condition/status of the event. This is actually more about "an expression of the event/trigger". How about we call it "EventExpression" or "triggerExpression"?

tsurdilo: Good point. We call the object type of this parameter "condition" but as you mentioned that is confusing. Will update here and also in error definition as its used there as well.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -476,7 +476,7 @@
     },
     "eventstate": {
       "type": "object",
-      "description": "This state is used to wait for events from event sources and then to invoke one or more functions to run in sequence or in parallel.",
+      "description": "This state is used to consume events from event sources and then invoke one or more actions to run in sequence or parallel",
       "properties": {
```
cathyhongzhang: Suggest to change to "This state is used to wait for events from event sources, then consumes events and invoke ...". I think it is important to let people know that when the workflow is at this state, it will wait there for external events to happen, and will not do any actions or transit to next state until that event or events come.

tsurdilo: Ok will update.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -502,12 +502,12 @@
         "filter": {
           "$ref": "#/definitions/filter"
         },
-        "eventActions": {
+        "consume": {
           "type": "array",
```
cathyhongzhang: I think it is better to call it eventActions. Action is more generic, which could be any function (some function could result in sending/outputting out another notification/event). I am not sure what you have in mind about event producer. From workflow spec point of view, workflow is a more an event consumer. It is the function/action that could produce/emit an event. wdyt?

workflow/spec/spec-examples.md
cathyhongzhang: Yes, fully agree that Conformance with this specification is on the workflow definition level. When the spec is more mature, we can have a document describing some example implementation design to help bridge the understanding gap.

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: i thought a lot about this and coded several examples using both. one thing that becomes confusing is that this state is already called an "Event State", so up-front its clear what it deals with consuming incoming cloud events. it makes it more readable with consume imo rather than re-instating further "event*" properties. 

Regarding consume/produce:
Functions as far as workflows are concerned are a black box (they can solve business requirements, which include producing other events, if needed). Workflows are responsible to configure how they are invoked, decide when they are invoked, what data to pass to them, and then to get their results data so further orchestration decisions can be made.

If workflows can be invoked via events, they should be able to produce events when they complete. This is important for event-driven orchestration. The completion of an orchestration workflow might need to trigger invocation of a subsequent orchestration workflow. 
Hope this helps

tsurdilo: @cathyhongzhang and btw if we do this (produce event on workflow execution completion) I am all for adding the end-state back as then it can be reused :)

cathyhongzhang:
It is an interesting idea that a workflow produces an event when the the workflow execution completes. I am just wondering in which scenario this will be used. Do you have a scenario in mind? If it is just for invoking a subsequent workflow, wouldn't the subflow state cover it already? My thought about so called "event producer" is that a user's function could do something like saving  data/document into a storage or sending an email out, and this action can be an event which triggers a separate workflow execution. The user's serverless function can do anything, which is out of the scope of this spec, and we do not reqirue the user's function to have any orchestration logic.

cathyhongzhang:
W.r.t. your comment "up-front its clear that it deals with consuming incoming cloud events. it makes it more readable with consume imo rather than re-instating further "event*" properties", since it is clear that it deals with consuming events, then we do not need to say "consume" again. This definition is the definition of the actual actions/functions that would be executed due to happening of the event.  So I think eventActions better matches the definition.

tsurdilo:
this is not new in worklfows, see for example bpmn2 message end event. 

its just another form of event, event that signifies the completion of that particular orchestration workflow. it can trigger other workflows or other event based services or be logged etc. many uses for it

cathyhongzhang:
I would suggest that you separate this PR into two. We can merge this update without the producer part. We need more discussion on the producer part since it is a new concept.

tsurdilo:
@cathyhongzhang 

regarding " So I think eventActions better matches the definition": yes I understand your point but the problem is the word "event" in "eventActions", it should be "events" and "eventsActions", to explain:

currently "eventActions" is a list of objects which define:
* an expression that resolves to a particular type of defined event
* a list of actions to be performed

so you can have multiple of these objects, basically saying in a single event state:
* on event A, perform actions 1-10
* on event B, perform actions 11-20
etc etc

"eventActions" sound as we only have a single event definition and perform multiple actions, but it is more like we can perform actions on multiple types of events. Does this make sense ?

tsurdilo:
@cathyhongzhang yes it will be two separate prs, I just brought up this conversation about produce to explain the naming of "consume" property, sry.

tsurdilo:
@cathyhongzhang wdyt about "eventsHandler"?

cathyhongzhang:
I got your point. How about we go with "eventsActions"? It is not an English word, but a combination of proper English words. We already have other primitive names which are combinations of English words.

cathyhongzhang:
"consume" is a verb, not a noun, which does not go well with other names which are nouns.

tsurdilo:
@cathyhongzhang ok "eventsActions" it is :) I will make the update.

tsurdilo:
@cathyhongzhang just one more - would "eventsHandler" work?

cathyhongzhang:
I like eventsActions better. I feel it is more about what actions to take when events come. Action seems more broad.

tsurdilo:
rebased and updated
@cathyhongzhang could you look again?

cathyhongzhang:
LGTM

PR #150, opened 2020-01-23
Removing loop parameter in operation state

tsurdilo:
Just small fix to remove loop parameter in operation state. It should not be there.


cathyhongzhang:
LGTM

PR #153, opened 2020-01-30
Enhance cloudevent matching in event states

tsurdilo:
* Aligns referencing events in event states with how referencing functions in actions is done
* Adds addition properties to event definitions which are properties of CloudEvent format
* Adds ability to do event matching in event states based on the CloudEvent payload (data property)
* Renames "functionref" to "functionRef" to align with "eventRef" (same with its "refname" property)


cathyhongzhang:
Maybe I misunderstand the change. I don't see a need to change the "expression". In many applications, an action is triggered by an"and/or" combination of events. We need to support this:  "Boolean expression which consists of one or more Event operands and the Boolean operators". Existing spec defines all the event attributes in the trigger event itself (if needed, we can add more attributes to the event definition itself for more granular definition of events), then in the eventState and eventAction or any other places that need to reference the event, we just reference the event name/ID.

tsurdilo:
@cathyhongzhang In your case then there is no need to define events at all as the expression can be evaluated against the incoming events (which it has to anyways). 

we can make refName a condition which can be evaluated against defined event names then you could do for example "event.name eq 'A' or event.name eq 'B'"

cathyhongzhang:
I don't think we can or should have a requirement : if you want to use the workflow spec, you must have a "Cloud Watch" like event processing rule engine to filter or generate another set of "orchestration events".  In all our customer cases, the actions are triggered by one CloudEvent or a simple "and/or/not" combination of the CloudEvents. I agree that as more applications start to use the workflow spec, there will be cases that need complex event rules and theoretically the rules can become very complex. Here is my thought: we can state in the spec that if the event rules are complicated, it is expected that the user has a middleware entity (like CLoud Watch) that will handle the complex event processing rule and generate a new CLoudEvent. This new CloudEvent will then be consumed by the workflow engine. But in majority cases, the original event producer can directly send the events to the workflow engine. The boolean expression of events works for both the majority cases and those few complex cases.

tsurdilo:
@cathyhongzhang yes the way we deal with events in this PR offers both. The way to consume all events from an event source with a certain type and other params, with additional check of the event payload.

In cases where the an event state needs complex interactions between events this is better suited for complex event processing services to define those rules.

For us to define an expression however is counter-intuitive as it really does not provide any help for users as it kinda sits in-between those two.

tsurdilo:
I will update the text to mention what you stated.

cathyhongzhang:
I am afraid that I can not agree with some of your points. Most, if not all, event Bridge/Grid/Router will do filtering based on CloudEvents params and event payloads, but will not do boolean combination of different events. So our workflow does not need to duplicate the work of filtering/checking on CLoudEvents params and/or payload. What we need to provide is the missing functionality of boolean combination of events.

tsurdilo:
updated

cathyhongzhang:
The boolean expression is not complex event rules. They are just boolean combination of event names/IDs, which is needed in many usage scenarios.

tsurdilo:
@cathyhongzhang I think even the smallest of expressions you might want to define like 

Event "A" and event "B" includes temporal reasoning which already is a complex rule. 
Event "A" or event "B" - again same thing..what is the duration, what is the type of our event streaming mode, do we have any "sliding windows" during which we define the existence of one event but not the other....
workflows are **not** responsible for this

tsurdilo:
@cathyhongzhang regarding the payload to explain the reasoning behind the addition:
Since our event state is **not** a switch state (we had these discussions iirc before), we can express currently:

"If an "Order" event  happens perform "Automatic Provisioning Actions""

but we cannot at the same time express:

"If an "Oder" event happens with cost > $1000 perform "Manager Approval" actions". If it's less than 1000  perform "Automatic Provisioning Actions"

The decision what actions to perform is **orchestration logic** and should be able to be handled by the workflows which are responsible for orchestration logic. 

The way we have to handle it now is pass the event payload to functions and let them decide what to do (manager approval or not), and functions are responsible for business logic not orchestration logic. Just my 2c : )

tsurdilo:
@salaboy @manuelstein @mswiderski - what do you guys think? I think this is maybe good discussion to include more people as it is a big part of the specification.

cathyhongzhang:
There is no temporal reasoning with "Event A and event B". It just means both events need to happen before the action is triggered. Same for "Event A or Event B", which just means etiher event could trigger the action. There are scenarios that need all the events to happen or just need one of them to happen. If the user needs temporal or other complicated relationship between these events, they need to have a separate event rule engine.

cathyhongzhang:
Payload varies a lot among different event sources/producers. I don't feel it is a good idea to bring the payload filtering into the event expression since it can make it very complicated. It is better to leave it to the middleware "event rule engine". Also as I mentioned before many existing middleware or event bridge/grid/route already support paylaod filtering and there is no need for the workflow to repeat this.

tsurdilo:
@cathyhongzhang 

(1) Our Event task defines a parameter:
**timeout** | Time period to wait for incoming events which match the expression
So from the start we define presence of events **in terms of time** in that state - "A and B within $timeout$"

(2) If we allow users to say "Event A and event B" with an expression you run into immediate problems of community wanting to specify "A before B" , "B before A", "A and B at same time" which you cannot use a boolean expression for. 
Do we allow "A and A"?
How do we deal with data integrity with "A and B and C" for example when we merge all events data into the state data input (overwrite)? Using boolean expressions quickly will become an implementation nightmare  as well as non-portable. 

Our spec has to be precise:
"If an instance of event A -> run these actions" "If an instance of event B -> run these actions"...that is portable. How these events A, B are produced, or if one cloud provider has CEP engine, but another has an event bridge, we do not care.

(3) **Payload varies a lot among different event sources/producers** -- thats why we enforce the CloudEvents format which lifts this ambiguity and  payload is the data element which can be of type "object" (JavaScript) or "string" as per that specification
What we are doing here is making orchestration decisions (what actions to execute) based on data.
If this data comes from and event payload from a CloudEvent, or since we define that event payload can be merged into a state data input (via filter) and then it is reasoned over by for example by next choice state **it is the same thing**. 

I think it would be very useful if you could describe how "event-expression" works in FunctionGraph. The description indicates to point to a name of a defined event, which is exact same thing as what we are trying to do here.

cathyhongzhang:
The timeout is outside the expression. It is a timeout for that state, which denotes that the workflow will transit from this event state to next state if the event expression is not satisfied within that timeout period (A side note on this: We can update this part to allow the event state to trasit to different next states when expression is satified vs timeout, wdyt?)  

I don't feel adding timing logic into the event expression, such as "event A before event B or event B happening X min after event A , etc.", is a good idea since it can become very complicated. 

BTW, the CloudEvents WG is discussing subsciprion/discovery API including filtering for CloudEvents and how an event pub/sub engine/middleware should work with CloudEvents producer on information passing and filtering. So I think we can probably leave this filtering of "external" CloudEvents to the pub/sub middleware or subscription API. As to filtering of information defined in the spec (here the information could be event, or function executuion result, or internal state parameter), I am seeing it as info internal to the workflow. 

I am not sure how latest Function Graph works. But the first release only supports the simple case, one event, with future extension to support event-expression (and/or/not combination of events ). 

I think we need more people's input/opinion on this. I think this will be a good topic for our first weekly meeting discussion.

tsurdilo:
@cathyhongzhang 
** The timeout is outside the expression. It is a timeout for that state, which denotes that the workflow will transit from this event state to next state if the event expression is not satisfied** 

Nope...it is currently part of the "eventsActions" array so the state can have multiple timeouts each associated with the **event expression**. It is very closely associated with the event expression and its definition related to **time** (in addition adding even more confusion along with the expression).
Look at the spec.md please first before making these statements because we are seriously wasting time here instead of making progress.

cathyhongzhang:
Can I request you not to make negative "waste time" comment? Negative comment is bad for community cooperation. 

Of course I read the spec before making the statement. You misunderstand my statement! If you read carefully, you can see what I said is "outside the expression", which does not mean "outside of eventActions".  The timeout is not associated with individual event inside the expression, as shown in your example.  The timeout is associated with the expression as a whole, which means that the workflow will transit from this event state to next state if the whole event expression is not satisfied within that timeout period.

tsurdilo:
@cathyhongzhang 
**I am not sure how latest Function Graph works. But the first release only supports the simple case, one event, with future extension to support event-expression (and/or/not combination of events ).**

And that's exactly how we should tackle this as well. Start small and well-defined and then in later versions and per community request add things as needed.

The current way we describe this is way too confusing and you run a high risk that it's not portable / vendor neutral at this point in time.

I would suggest going with these changes in addition to making the timeout property be on the state level as a fallback-all option.

tsurdilo:
**The timeout is associated with the expression as a whole, which means that the workflow will transit from this event state to next state if the whole event expression is not satisfied within that timeout period.** 

so you have 
"eventExpression": "event.name = 'A'"
"timeOut": "PT2S"

and I have 
"eventTimeExpression: "event.name = 'A' and seconds <= 2" or something...what is the difference
2 properties instead of 1 to define event existence in terms of time...

cathyhongzhang:
Compared with some new additions, I think the "and/or/not" is small and well defined. Let's get more input from other people in our weekly meeting. We need more people to comment on many other parts of our spec. It is not good to make this spec a one or two people spec.

cathyhongzhang:
For your single event expression example, there is not much difference, but for an expression consisting of combination of events, there is a difference. For example, if the expression is "event A and event B" , timeout is 60s, it means the workflow will only execute the action in this event state if  both eventA and eventB happen within 60s (it does no matter eventA happens 2s or 20s ... before/after eventB).

tsurdilo:
@cathyhongzhang sorry I will just add one more thing, maybe discuss it at the meeting please:
from your example "event A and event B":
A = Normal Temparature Event
B = High Temperature Even
timeout 60s

if A comes before B we might want to run the "call the doctor" actions
if B comes before A we might want to "go outside and play cause fever is over" actions

boolean expression is not optimal for this imo.

salaboy:
@tsurdilo @cathyhongzhang very interesting conversation, I tend to agree with @tsurdilo because we come from the same background. In general, I would be interested in seeing some very simple bindings with Cloud Events, so basically define that the workflow can receive and emit events that are relevant for the application (not workflow events). 

If you go to all the expressions handling and event correlation it gets quite difficult to handle and there are already other specs that are trying to solve these expression problems.  
When I look at the definition of the issue by @tsurdilo it is quite clear what the intention is, when I think about @cathyhongzhang suggestions about adding complex expressions then it becomes fuzzy and it open the door for way more complex requirements that I don't think we should target in this spec.

cathyhongzhang:
@tsurdilo w.r.t the following: 
if A comes before B we might want to run the "call the doctor" actions
if B comes before A we might want to run the "go outside and play cause fever is over" actions

As you can see, if we bring the sequence or timing into the expression, it could become very complicated. So my point is not to bring inter-event sequence or timing into the boolean expression. The boolean expression is just a simple and/or/not combination of events (without any inter-event timing/sequence) which as a whole will trigger some action. If some application involves complicated inter-event timing/sequence as the action trigger, then a CloudEvent bridge/grid/router should be used, as we agree before.

cathyhongzhang:
@salaboy Yes, I agree with you that we should not add complex event expression. I am not suggesting adding complex event expression. With so many comments going forward and backward, I understand it could be confusing to you. Let's discuss this in the weekly meeting for a concensus before merging this.

tsurdilo:
@cathyhongzhang I think the confusion comes from what you mention 
like simple and/or/not combination etc. - could you please provide a real-world example it would really help.

cathyhongzhang:
@tsurdilo Let's take a surgery application workflow as an example. The patient is being monitored for multiple physical information. Each monitoring device produces an event and there are multiple monitoring devices. If any one of the information exceeds some threshold, an alert should be raised. So this needs an "or event" expression. 

In a college application workflow, multiple events need to happen before the application deadline in order for a college to mark your application as "completed": the student needs to submit the application, the SAT board needs to send the student's SAT score to the university, two recommendations need to be sent from two individual teachers, etc. So this needs an "end event" expression.

duglin:
Still coming up to speed on things, so sorry if I'm misunderstanding things, but it seem like part of the difference of opinion here is concerned with what's currently in the spec with respect to the boolean expression support. And from my reading of the spec, while it seems easy to express something like "if event A and event B happen, then do ...",  actually implementing this could be quite hard when you consider interoperability and portability.

I say this because, as @tsurdilo mentioned, it seems like there are many aspects that are left as an exercise for the developer and therefore if people make different decisions on those aspects then you'll get different semantics on different implementations of the spec. For example, how long should the impl wait for "B" once it saw "A"? What if 2 A's were seen, which one should be used when "B" is seen? Or does the expression result in being "true" twice?

I don't think any of these questions are insurmountable to define, but they are not currently defined in the spec but defining them might take some time to get right.

Would it make sense to split this concern into two steps so that progress can be made?
Step 1 - take this PR as is so that a) the other good changes can go in, and b) it removes an interop issue for now
Step 2 - open an issue as a nagging reminder to revisit the boolean expression support and it gives the people who want that support to be native in the spec (instead of moved into a service) time to fully flesh out the idea

cathyhongzhang:
@duglin  I think we already agree not to bring the inter-event sequence or timing into the expression of the workflow. If a user's scenario involves complicated inter-event sequence or timing, such as "how long to wait for B once it saw A" as shown in your example or other aspects mentioned by Tihomir (it could become very complicated), an intermedite event GW which handles those complication needs to be used. We only allow the user to specify simple logical "and/or/not" operation on the events or just a single event in the expression, which should cover majority use cases. 

Just to calrify again: if the user specifies "event A and event B", it just means both events need to happen with no timing/sequence relationship between them. 

I do not see any interop issue since we are not targeting at interop with any specific vendor's implementation.

duglin:
@cathyhongzhang if "A" happened 2 years ago, and "B" happened today, would that satisfy the boolean expression?  I can see two devs answering it differently. One might say "of course since there was no time limit specified", but another dev might say "heck no, I'm not going to hold on to the info for 2 years".  That's the type of interop problem I see and so I think it would be best to try to split this situation into two separate concerns so one doesn't block the other.

cathyhongzhang:
@duglin Good question. That is why there is a "timeout" associated with the expression as a whole, which allows the workflow application to specify how long the workflow engine should wait for the expression (both A and B must happen) to be satisfied (if the timeout is not specified, then it means that the application requires the workflow engine to wait forever). So there is no ambiguity and the interop issue you described will not happen.

tsurdilo:
timeout is not associated with the timestamp property in the event data. Its as defined associated with the amount of time once event state becomes active to wait for existence of events. An event that happened 2 years ago as mentioned can still exist so i think dugs case still valid independent of the timeout property. 
 
In this pr we are adding optional matching against event payload so we can cover cases where the timestamp prop in event data could be checked. using expressions with unions and intersections of events makes that ambiguous as far as impls go. 

Also we never addressed the state data issue when you have lets say "a and b and c ..." where we currently define that event data is merged into the state data. You are there opening a can of worms with merging, overlapping, and maintaining the integrity of your state data. 

The use cases you mention that expression cover are small and do not really portray real world problems. The amount of issues it brings compared to the use cases it covers is simply not worth it imo.  But still, without using expressions users can model their workflows to cover your use case ("and" as simple transition from one event state to next, "or" as event state inside parallel state) but without all the many issues we mentioned in this thread that expressions can cause.

salaboy:
@tsurdilo @cathyhongzhang @duglin 
100% agree with Doug's comment and suggested steps:
```
Step 1 - take this PR as is so that a) the other good changes can go in, and b) it removes an interop issue for now
Step 2 - open an issue as a nagging reminder to revisit the boolean expression support and it gives the people who want that support to be native in the spec (instead of moved into a service) time to fully flesh out the idea
```
For step 1 I am not sure why we need more discussion.

manuelstein:
I share that understanding of "timeout" - it starts when the workflow enters the Event-type state.

> @salaboy @manuelstein @mswiderski - what do you guys think? I think this is maybe good discussion to include more people as events are a big part of the specification.
I'm obviously not a fan of the existing spec and Event-type state plays a good part in that but let me see. So, the Event type was meant to be able to consume more than one event at a time. Unfortunately, the boolean logic (expression) that was meant to correlate such independent events is unfit to express causality or timeliness and hence causes ambiguity.

Not merging this PR means keeping the arbitrary boolean expression - knowing that it can result in ambiguous workflow definitions. @duglin asks if it was okay to merge this PR for now (remove boolean logic) but to start working on a solution immediately to combine events.
+1 on this

I think it's a discussion bound to happen because the problem goes beyond the Event state. The correlationToken is a great thing because one can send events to an already running workflow instance. But only Event-type states can consume those events for now. It's unclear what happens to events that arrive while the workflow instance (activation) is not in an event state, e.g. where to deliver them and for how long events should be queued or where.
Desired application behaviour could be described in a discrete event dynamic system model (e.g. Petri), but that's probably overkill model complexity for most of the use cases, so IMHO we should possibly just specify what happens when an event with the same correlation token appears.

This requires a few answers to questions like:
- should a workflow have a single startAt or one per trigger/subscription? i.e. can there be concurrent processing/flows under the same workflow activation?
- Or do multiple events just get injected/merged/batched into the ongoing workflow (e.g. nested events? ;-) @deissnerk)
- could an incoming event interrupt a running action?
So, also +1 to discuss the overall issue in a call, which might also provide a solution on how to combine independently arriving events.

tsurdilo:
@manuelstein : **It's unclear what happens to events that arrive while the workflow instance (activation) is not in an event state, e.g. where to deliver them and for how long events should be queued or where.**

This can differ per implementation but I'll try to explain the general approach with example workflow:

Screen Shot 2020-02-05 at 11 13 17 AM

Let's say "Event State 1" actions are triggered by CE "A" and "Event State 2" for CE "B". 

And lets say the order of events is AABB

First "A" would trigger one instance of workflow execution. It would execute actions in event state 1 and transition to event state 2. At this point workflow is again in a wait state. 
For the second "A" there are two options - it either has a correlation token that correlates to the first workflow instance or it doesnt. If it does, its correlated to first process instance which is in the wait state for a "B" event, so nothing is triggered.
If the second "A" does not have a correlation token, this will invoke a second instance of the workflow, again performing actions of event state 1 and moving to event state 2.

For the next 2 events "B" "B" lets say first "B" has correlation token for first workflow instance, that will trigger event state 2 and workflow ends. If it does not have a correlation token and there is no workflow that is triggered by event "B" it is ignored. Same thing for the second "B" basically if it has a correlation token for a workflow instance it is consumed only if that workflow instance is in a wait state which waits for that event type. 

Hope this helps.

tsurdilo:
@manuelstein: **should a workflow have a single startAt or one per trigger/subscription? i.e. can there be concurrent processing/flows under the same workflow activation?**
Depends on how many "eventsActions" objects the event state has and if it is a startsAt state or not.  If you have a "startsAt" event state which declares lets say two eventsActions, one for event A and the other for event B you can visualize that as:

Screen Shot 2020-02-05 at 11 46 18 AM

**could an incoming event interrupt a running action**
No. The workflow instance has to be in a event/wait state which is actively looking to consume it.

manuelstein:
Thanks, so this example defines that there is no queuing/buffering and events would be ignored unless the workflow is in a fitting Event state.

1) What if AB arrive in order only that B arrives while the workflow is still processing event A - then I assume B would also be ignored because the workflow has not yet transitioned to State 2?

I guess something must have started the whole workflow before it has reached state 1 to consume an A event, so I assume there was also some event X with the same correlationToken that started the instance.

2) What would the workflow describe as triggers? If it subscribes to [X,A,B] I would expect that any of the three types could trigger a run of the workflow starting with "startAt". It could be only subscribing to [X] and when entering an Event-type state; the state would subscribe to the types/sources it waits on and unsubscribe once it is leaving the Event-type state.

In your example that would be e.g. WF.events = [X], when our instance with a specific correlationToken enters State 1, it subscribes to [A] (maybe with filter on correlationToken) and transitioning to State 2 it unsubscribes from A and subscribes to [B] (again only for a specific correlationToken) . Does that sound okay? It's very consistent, but it might not be what users expect, especially when actions/events take more time than expected.

tsurdilo:
@manuelstein:
1. This goes back to why we are kinda doing this change :) Workflows define the repeatable logical steps for orchestrating business logic. What we are getting into are services for complex event processing which we cannot describe. Yes there has to be services to start workflow execution (maybe thats done on deployment,  a rest call from a function that triggers an orchestration workflow, or in case of event startsAt state impls might start workflow instance on events...thats not up to us imo). 
2. Yes if i understand correctly.  Note that triggers can also be user interactions/decisions (which hopefully we can define in spec at some point as well).

manuelstein:
To the second example - this means that IF the state that startsAt points to is also an Event-type state, then the event that triggered the workflow invocation will also be consumed by the Event state, but in any other case the Event state would wait on external arrival of a matching event?

So, looking back at the A-B-example: Event A triggers the workflow, is consumed in State 1 and some processing happens, some result JSON or output is being generated and the workflow transitions to State 2. If the output of State 1 matches a type-B-event (i.e. what State 2 wants to consume), then State 2 continues processing. Otherwise it would discard that input completely and wait on the arrival of a type B event, correct?

That is one working assumption, although a little bit confusing given the WF's events attribute that lists what can trigger a WF execution and the startsAt where every execution would start. Maybe the startsAt could be given up completely in favor of one or more Event-type states (or Wait states as they're waiting on signals).

tsurdilo:
@manuelstein: I would say yes, if workflow instance is in an intermediate wait state (process instance already stated) the consumed events must include correlation token. 

I dont know if fully understand **If the output of State 1 matches a type-B-event** Think at this point we are only talking about external events.
The specification defines that the data output of state 1 becomes the data input of state 2. State 2 then would have this data available to use, so when lets say external event B happens with correlation token, it could use this json data to pass as parameters to functions of that event state 2 actions. does this make sense?

We still have cases where workflow execution might not need external events, so startsAt is useful,but thats definitely something to think about imo.

cathyhongzhang:
I like to clarify one more time that w.r.t Doug's example, existing event boolean expression has no ambiguity. If the expression is defined as "event A and B" and the timeout is defined as "10 min" in the event state, then the implementation should wait for 10 min for the two events to happen. It does not matter which event happens first or what is the time gap between these two events. If both A and B happen within 10 min, execute the action. On the other hand, if the expression is "event A and B" and there is no timeout defined in the event state, then the implementation should wait forever for the two events to happen. 

Existing boolean expression does not and is not intended to support complicated timing realtionship between the events,  such as "A happens before B or B happens 5 min after A". I am not sure whether we should add/pull the definition of those complicated event filtering or inter-event timing/sequence, etc. into this spec because it could become very complicated. 

I think the confusion about the and/or/not expression is because people assume there is implicit sequence/timing between the events and thus feel it is ambiguous. From all these questions, I think we need more description/clarification on the semantics and usage of each definition. Of course this spec is far from perfect, it is still in very early draft stage. 

If I understand correctly, this PR adds event payload filtering into the event trigger/expression, e.g. "event type A with payload field X =a or payload field Y=b but payload field Z!=c ".  
CloudEvents spec is defining subscription/discovery API which covers the the event payload filtering. Why not leverage that instead of defining our own in this spec? Payload filtering could become tricky and complicated.

cathyhongzhang:
@manuelstein You raise some very good questions. We can discuss them in the weekly meetings.  
W.r.t. your point "it might not be what users expect, especially when actions/events take more time than expected", I agree. I think we can not reach conclusion that there is no queuing or buffering of events.

cathyhongzhang:
As to how the workflow is invoked, my thought is that if the "startAt" state is an event state, then the workflow can be invoked upon receiving the events defined in the event state. Otherwise, the workflow can be invoked by an external API call or CLI. There is once a suggestion that we enforce the workflow to start at a event state since an API call or CLI can also be modeled as a trigger event. Maybe we can take that suggestion?

tsurdilo:
I'm confused as to what is going on here after all these discussions. 
Just looking from the comments  we have 4 members agreeing to move forward with these changes and 1 against. 

@duglin what do you suggest is best action to proceed with this at this point?

mswiderski:
A bit late to the party but just to add few thoughts from my end... 
so first of all as soon as we are talking about events with relationship between each other (and/or/not) we directly include timing in the mix. Why is that? It purely because these events must be correlated - there is no other way to ensure workflow will start without taking all events into consideration. So let’s take them one by one starting with the easiest

- **or** - this is the simplest case as it defines that either event A or event B will start new workflow instance. So that means in most of the cases is just an alternative ways of starting workflow instances.

- **not** - this gets more complex because it implicitly means it is relative in time or other events. Otherwise such workflow would be started for any other event than the one nagated 

- **and** - this in my opinion always means time dependency between events and at the same time enforces correlation token/key as this must be used to know if existing workflow instance should be used or a new instance should be created. Correlation token/key in ideal case should be business driven such as part of the event instead of identifiers generated by the workflow.

So attempting to use just boolean expression for event state definitions will be rather complex and will make implementations to differ in behaviour. At the same time combination of event states and switch states should provide similar, though explicit support for exactly that requirement.

cathyhongzhang:
Yes, agree that combination of event states and switch states should provide similar support for the requirement. But isn't this like pushing the complexity to the workflow application user? Let's take this requirement as an example: event A and "eventB or eventC". If we ask the user to specify this requirement via flow diagrams, the user needs to provide quite a lot flow diagrams. Is it better that we push this complexity to the user or should the workflow engine implementation translate the expression into flow logic internally and take over the complexity from the user?

cathyhongzhang:
I understadn the implementation could be complicated. The question is whether we should push this complication to the user or the workflow enginer should deal with it. I am open with taking the expression out, but the complexity, especially if we include timing into the mix, is still there. Someone or some entity needs to take care of this complexity.

cathyhongzhang:
100% agree with "and - this in my opinion always means time dependency between events and at the same time enforces correlation token/key as this must be used to know if existing workflow instance should be used or a new instance should be created. Correlation token/key in ideal case should be business driven such as part of the event instead of identifiers generated by the workflow."  I raised the "correlation token" discussion in CloudEvents back in March 2018, which was not understood initially by many people as to its need. Good thing is that the CloudEvents team is a very open and healthy community. People are willing to spend  time understanding its use cases and discussing how to support it.

tsurdilo:
@cathyhongzhang I think the timing aspect is looked at at comment https://github.com/cncf/wg-serverless/pull/153#issuecomment-582491948 but we can go into some more specifics if needed.

mswiderski:
@cathyhongzhang I'd say it's less about pushing the complexity to workflow users but actually giving the explicit control over what happens. Attempting to hide it with boolean expression would make more issues for the workflow users than the explicit modelling. 

What's even more important in my opinion is to allow users to have a clear need to make such event correlation in the first place. As I agree with comments made before that such event correlation/merging/aggregation is not workflow responsibility. So better to have initially supported more explicit way of defining things and start "hiding" more based on feedback from both workflow users and vendors.

cathyhongzhang:
There seems to be no difference w.r.t. giving explicit control to the user whether using a boolean expression or using multiple flow diagrams. At high level, we can divide the use cases into three categories: 1. the use case does not involve any inter-event relationship, i.e. just one event 2. the use case involves event union and/or intersection relationship but no inter-event temporal relationship. 3. the use case involves inter-event temporal relationship. Case 1 is simple. For case 2, isn't it better for the user to just specify the use case via a simple event expression in one workflow diagram instead of asking the user to specify the use case via multiple workflow diagrams?  As to case 3, since case 3 can become very complicated, we can leave it outside the workflow engine implementation. Either way has its pros and cons. It all comes to where is the balance point. I am with you about starting simple and adding more support based on feedback from both workflow users and vendors.  We have spent enough time on this. Let's discuss this and reach a conclusion in tomorrow's meeting.

salaboy:
@mswiderski that definitely brings clarity on why this is difficult. So my gut feeling is to make sure that we don't drag that complexity right now. If we say, let's start with OR, we open the door to support NOT and AND later on. Starting with OR makes sense to me, but we need to be clear that this is what we are willing to support, for now, to avoid people asking for AND and NOT.

mswiderski:
@cathyhongzhang I don't think this use case is valid

> 2. the use case involves event union and/or intersection relationship but no inter-event temporal relationship

as soon as you define that workflow should be triggered when two events are received they must be time tracked so to say.

@salaboy I agree here, let's start with just OR use case to keep it simple for now and see how this evolves and what needs will arise.

duglin:
If there's consensus that other operands beyond OR would head down a path that people want to avoid, would it make sense to think of it as a list of "events" (meaning an array where any event in the list triggers it) instead of an "expression" ?

cathyhongzhang:
@mswiderski , there are use cases that an action is triggered when multiple events happen before a time point but they do not care whether one event happens before/after another event (no inter-event temporal relationship).

tsurdilo:
@cathyhongzhang I think might be confusing facts with events. Facts do not have a time constraint, events always do. 
Multiple facts lets say like "Applicant age is over 18", "Applicant address is in Germany" can be reasoned over to deduct an event such as "Application Granted"
But events, like "Application Submitted" and "Application Granted" **always** have a temporal (time) relationship.

This is I think what we have been trying to convey here from the beginning.

cathyhongzhang:
@tsurdilo You seem to misinterprete the college app example I gave. In this example, multiple events refer to the following:  eventA--math/science teacher's recommendation is received, eventB--language teacher's recommendation is received, eventC--SAT score is received. It does not matter whether one event is received before/after another event. As long as they are received before the deadline, an action will be triggered upon receiving all.

cathyhongzhang:
Maybe we can start with support of "just or events expression" and "just and events expression", but not any combination.

tsurdilo:
I think we have several people here clearly stating the pitfalls of using expressions to begin with and their agreement not to use them in this particular case. Or am I missing something? 

This pr is the starting point so we can then improve to support more stuff based on discussions.

tsurdilo:
i think still you have then a temporal relationship of 
two events that must be **between** "Application Started" and "Deadline Passed"

cathyhongzhang:
@tsurdilo You are talking about different things. What I am talking about is that there is no inter-event temporal relationship between THESE three events in the exmaple I gave, and the user can just use one expression instead of three flow diagrams to specify it.

mswiderski:
@cathyhongzhang you're right that there is no inter-event relationship (or it does not have to be) but as soon you need to wait for more than one event how do you know if these events should belong to same workflow instance? This is what @duglin has already brought up earlier in the discussion. 
In addition as soon as first event arrives then the workflow needs to consume that event and as such start new instance and wait for the other one and then somehow correlate these events where time round should be in most of the cases assigned as well. And this is the main issue - time assigned to give up waiting and correlation is a must to make this usable at all. 

Having single expression to cover all is just too generic to address at first thus suggestion to start with OR only makes the most sense in my opinion.

cathyhongzhang:
@mswiderski, If I understand you correctly, you are concerned about how the workflow engine knows whether these events should belong to same workflow instance or not.  Good question! 

To handle this, we need the correlationToken and the timeout defined in current spec.  Let's use the college application use case as an example, i.e. some action should be triggered when eventA, eventB, and eventC are all received before some application deadline. Suppose the user defines appplicant's ID as the correlationToken and the deadline as Dec 31. The following is one possible way to do it: when the workflow engine receives any of the three events, it will check whether there is already a workflow instance that is associated with the correlationToken carried in the event msg. If not, a new workflow instance will be created to handle this event and this new workflow instance is associated with this new correlationToken, and a timer will be started. If yes, the event will be delivered to the existing workflow instance. If the timer times out before the workflow engine receives all three events, the workflow engine will transit to the end state or another state specified by the user. If the last event is received before timeout, the workflow engine should trigger the action. Duplicate events should be discarded (we can discuss how to indentify a duplicate event if needed). Different use cases will most probably have different correlationToken. Each user knows its use case the best and it is up to the user to choose a proper correlationToken for its use case. It is also up to the user to define the event source and type. 

If we merge this PR which takes out the "and" expression of events (it also takes out the "or" expression), how will the user specify this: "eventA and eventB and eventC"? According to Tihomir's idea, the user can specify this expression using several flow diagrams, which is more complicated to the user than a simple "and" boolean expression. Furthermore, the main issue " time assigned to give up waiting and correlation", which you are concerned about, does not seem to go away and still needs to be handled by the work engine implementation. The duplicate event issue does not go away either.

tsurdilo:
**According to Tihomir's idea, the user can specify this expression using several flow diagrams**
@cathyhongzhang idk where anyone said several workflow diagrams? I said control flow logic and showed and provided examples of "and" being two connected event states.

The diagram I drew with multiple start states, its a visualization for "or", its 1 diagram not 2 or several.

Please let's not confuse people by saying some stuff that was never mentioned. 

So by taking this pr we really do not lose anything we didnt already have + we get rid of the expression string which I think by now everyone agrees to do but yourself, and then we start working from there to improve and add the rest of stuff we need.

tsurdilo:
@cathyhongzhang we will work on improving the event state to include all the possible cases, but we have to have a starting point and I think starting with reference to defined events is a good start so we can improve and implement all the other options you mention. This is not a final change by any means and please create issues as soon as this is done so we make sure its high priority.

cathyhongzhang:
@tsurdilo Sorry that I misunderstand you. Do you mean that you will combine those diagrams you drew, each with its own start state, into one diagram? Then the combined diagram would involve multiple event states (multiple start states) and maybe switch states, right? I may misunderstand you again. Could you please illustrate how you combine them by showing what the 1 diagram looks like for the example use case (i.e. some action "func" should be triggered when eventA, eventB, and eventC are all received before some application deadline)?

cathyhongzhang:
@duglin Yes, I am with you. There are multiple ways to express the "event or":  as a list of "events", using "or" operator, using "|", etc. Let's discuss what is the best way to specify this in the workflow meeting. We need to choose a way that can be consistently extended in the future to cover other possible cases.

duglin:
the advantage of an array is that there's no parsing, no expression language and people won't naturally think.... hey, let's also support this other syntax option

cathyhongzhang:
@tsurdilo I am not disagreeing with modifying the event expression to make it simple for workflow engine implementation. What I do not agree is the way your PR modifies it. I think we should allow people to discuss, clarify, and resolve all the questions and confusions so that everyone is on the same page. It is better to get input from more people in our next week's meeting about this design.

cathyhongzhang:
@duglin Yes, I see your point:-)

tsurdilo:
After all the comments in this PR I think I have a pretty good idea on how the event state needs to be changed in order to support "and" and "or" without using string expression. 

Thank you all for the valuable comments, I will close this PR and create a new one tomorrow.

Issue #155, opened 2020-02-06
Unmanageable event expressions in event states


tsurdilo:
This is related to pr https://github.com/cncf/wg-serverless/pull/153
This issue tackles one particular problem with current specification and lets call it

**Unmanageable event expressions in event states**

1. First lets look at parameters in a Cloud Event (reason being that our workflow spec is said to "consume" CE so our expressions must match against the properties defined):
* id
* source
* specversion
* type
* datacontenttype
* dataschema
* subject
* time
* data
* data_base64
(https://github.com/cloudevents/spec/blob/master/spec.json)

In order for a workflow then to say that event X might trigger some actions to be executed, it may have to match against these parameters. It may match against one, or all of them if needed. 

2. Lets now take a look at the event state as it is currently defined in our specification, here is an example:

```json
"states":[  
  {  
     "name":"Sample Event State",
     "type":"EVENT",
     "eventsActions": [
      {
         "expression": {
           "language": "THE EXPRESSION LANGUAGE USED",
           "body": "THIS IS THE EVENT EXPRESSION HERE!!!!!"
         },
         "actions":[  
            {  
               .... 
            }
         ]
     }],
     .....
  }
]
```
Event states can have one or more "eventsActions" elements each one containing an expression which needs to match events. 
The definition of this "expression" is: Boolean expression which consists of one or more Event operands and the Boolean operators. 

What are these "event operands" - that is not defined but in order to write an expression there has to be some reference to "$event1", "$event2", ......

3. Not let's attempt to write an expression, in our example lets say we are concerned only about one single CE:

```json
{
    "specversion" : "1.0",
    "type" : "com.github.pull.create",
    "source" : "https://github.com/cloudevents/spec/pull",
    "subject" : "123",
    "id" : "A234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "datacontenttype" : "text/xml",
    "data" : "<much wow=\"xml\"/>"
}
```

using lets say the popular Spring expression language and again assuming that an impl gives us reference to "$event"
we could write (now we always want to write "worst-case scenarion where we have to match on all properties so we can see what users might be having to do also):

```text
"$event.specversion eq '1.0' and $event.type eq 'com.github.pull.create' and $event.source eq 'https://github.com/cloudevents/spec/pull' and $event.subject eq '123' and $event.datacontenttype eq 'text/xml'"
```

[...]

PR #162, opened 2020-02-15
Event State Updates

tsurdilo:
Updates to event state based on:
* comments on https://github.com/cncf/wg-serverless/pull/153
* comments on https://github.com/cncf/wg-serverless/issues/155

Replaced string-based event-expression with a portable reference to defined events. 
Still maintain ability to specify multiple events (both "and" and "or").
Added capability to associate events for workflow instantiation (when correlation token is not available).
Added two examples showing off event state with multiple events (both exclusive and not exclusive).

You can see the spec updates with images all parsed here: https://github.com/tsurdilo/wg-serverless/blob/eventstateupdate3/workflow/spec/spec.md#Event-State
You can see the added examples with images all parsed here: 
https://github.com/tsurdilo/wg-serverless/blob/eventstateupdate3/workflow/spec/spec-examples.md#Monitor-Patient-Vital-Signs-Example
https://github.com/tsurdilo/wg-serverless/blob/eventstateupdate3/workflow/spec/spec-examples.md#Finalize-College-Application-Example


workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -471,13 +463,28 @@
             "$ref": "#/definitions/eventactions"
           }
         },
+        "timeout": {
+          "type": "string",
+          "description": "Time period to wait for incoming events (ISO 8601 format)"
```
duglin: Does this need to be specified as a duration rather than an absolute time?

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -456,12 +446,14 @@
           ],
           "description": "State type"
         },
-        "end": {
-          "$ref": "#/definitions/end",
-          "description": "State end definition"
+        "exclusive": {
+          "type": "boolean",
+          "default": true,
+          "description": "If true consuming one of the defined events causes its associated actions to be performed. If false all of the defined events must be consumed in order for actions to be performed"
```
duglin: in the "true"/or case, while I think it's implied, I think it might be best to be more explicit and add something about how if multiple events that match the array are seen at the same time, then each event would cause the actions to be performed independently and they are not grouped together into one workflow execution - unlike the "and" case.  That is correct, right?

workflow/spec/schema/serverless-workflow-schema-01.json
duglin: This is actually the time period to wait for the 2nd+ events in the "and" case, right? in the "or" case this timeout would have not effect I assume. I so, it might be good to state that the start of the timeout is when the first event is seen.

tsurdilo: > Does this need to be specified as a duration rather than an absolute time?

ISO 8601 format is good here imo.

tsurdilo: > This is actually the time period to wait for the 2nd+ events in the "and" case, right? in the "or" case this timeout would have not effect I assume. I so, it might be good to state that the start of the timeout is when the first event is seen.

Good point. I added a section describing the timeout property in event state. You can see it also here: https://github.com/tsurdilo/wg-serverless/blob/eventstateupdate3/workflow/spec/spec.md#eventstate-timeout

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: You are correct if event state is the starting state. Added more description under the two images showing the exclusive property in spec.md.

workflow/spec/schema/serverless-workflow-schema-01.json
duglin: What I was more wondering about is whether an absolute time (eg. `2020-02-18T11:39:27+00:00`) makes sense? I suppose someone could use it but it would seem to be pretty limited in value. Perhaps the spec should say that while any valid ISO 8601 value is allow, it is expected to be a duration for most cases - just to provide a bit of guidance for how we expect that field to be used.  Not a biggie.

tsurdilo: Added "Timeout values should always be represented as durations and not as time/repeating intervals." to explain that better.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -476,12 +466,14 @@
...
         },
-        "stateDataFilter": {
-          "$ref": "#/definitions/statedatafilter"
+        "payloadCorrelationKeys": {
+          "type": "array",
+          "description": "Elements of events payload used for defining correlation between incoming events. Can be used with event correlationToken or in case correlationToken is not defined"
```
manuelstein: A workflow's event definitions can specify a correlationToken for each incoming event definition. IIUC only events with matching correlationTokens would appear under the same workflow activation. In other words, the event state can only see events that correlate under the correlationToken. Why would two events that correlate to the same workflow activation require additional payloadCorrelationKeys in the event state and what for? If multiple fields are required to correlate events, why not solve this in the eventdef?

workflow/spec/schema/serverless-workflow-schema-01.json
manuelstein: IIUC exclusive defines the behaviour for each eventaction definition, correct?
If I have two eventaction definitions in eventsActions and the first one references events **A,B** but the second one references **A,C** - then **exclusive=true** means that either the first one will be executed when **A and B** have been received **or** the second executes if **A and C** have been received. But if **exclusive=false** then receiving A will cause both to execute, right? Is there a case where I want both eventaction definitions to be executed so when receiving A->B->C it would start the first one with A,B and also wait for C to start A,C until all eventaction definitions have been executed?

In another scenario, what if I have eventaction one depending on **A or B** and eventaction two depending on **C and D** ?

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: Correlation token is typically generated once a workflow instance is created. It can be the workflow instance id for example. In the case where event state is a starting state of the workflow, it defines the definition on how to start the process instance, however at the time events arrive, there is no process instance created, so there is no correlation token. 
For that case events must be correlated using their parameters/payload, for example associating incoming events that have the same studentid as shown in one of the examples. 

I do think that we probably do not need to set an explicit correlationToken parameter in the event definition, as the impls should set that once they create a workflow instance. 
Let's take that discussion to be part of another pr if possible. wdyt?

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: **IIUC exclusive defines the behaviour for each eventaction definition, correct?**
exclusive=true --> "OR"
exclusive=false --> "AND"

tsurdilo: Regarding your scenarios I will go into each one here in a minute but just to say that the only way to allow event state to deal with both "or" and "and" (exclusive/not exclusive) cases was to allow multiple events to be referenced by a single event action. Otherwise we would need a two separate states, an "exclusive" event state and a "parallel" event state.

tsurdilo: Regarding: 
**If I have two eventaction definitions in eventsActions and the first one references events A,B but the second one references A,C - then exclusive=false means that either the first one will be executed when A and B have been received or the second executes if A and C have been received.**

I don't think that is the case. We have to set some "common practices" here. If you have exclusive=false, it is best practice to define a single eventsActions. In your case however, you have to see multiple eventsActions as a combination, so having one for "A,B" and the second one "B,C", it really means that (since exclusive=false) that **ALL** events A,B,C are needed to start the workflow instance, so you can visualize it as:

Screen Shot 2020-02-19 at 10 04 46 AM

Where the first set of actions executed comes from the first event definitions and the second from the second. Hope this makes sense.

tsurdilo: Regarding:
**But if exclusive=true then receiving A will cause both to execute, right?** (for the same scenario as above)

Again the "common practice" for when exclusive= true is to reference a single event per events actions. However in your scenario again you have to see it as a combination for the particular case and can visualize it as:

Screen Shot 2020-02-19 at 10 14 40 AM

tsurdilo: Regarding:
**Is there a case where I want both eventaction definitions to be executed so when receiving A->B->C it would start the first one with A,B and also wait for C to start A,C until all eventaction definitions have been executed?**

This looks to me as the case for exclusive=false (you need all 3 events are needed) and a single eventsAction with multiple actions, so yes its supported with this pr

manuelstein: I also thought about this way to realize the cases. But depending on how I specify the events to be merged in exclusive=false, I might want the input to A^B would be different from the input to A^C. With exclusive=true, I rather figured the first and second set of actions to be like branches of a parallel state. (A circumstance that IMHO shows how overloaded and underspecified the functionality of the event state is)

workflow/spec/schema/serverless-workflow-schema-01.json
manuelstein: No, I don't think a correlationToken can be created at runtime. Two events either have the same value identified by their individual correlationToken path or not. If one event does not carry a value in the identified correlationToken that is more of an error case in which the workflow can't be executed (unless nil/Null counts as a value).

In case of the studentId it could be that event type A carries it in $.data.student and event type B has it in Microsoft's $.tenantId, type C could carry it in the OpenStack audit event's $.initiator.name - I think the current workflow's event definition provides that.

But if events need to concatenate/format/calculate a correlation token out of multiple attributes to identify the workflow they belong to, then this might need some more consideration, a union of fields identified by an array of keys is a start, I guess.

I thought the correlation should happen in one place, but of course, each event state could make different use of the events that the workflow is subscribed to.

manuelstein: this might be worth a separate discussion :) +1

workflow/spec/schema/serverless-workflow-schema-01.json
manuelstein: LGTM, +1 for this definition of exclusive

tsurdilo:
updated to remove the not needed "payloadCorrelationKeys" property and updated the definition of correlationToken with examples. 
@mswiderski would you mind taking a look when you have time? Thanks.

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: I updated the pr and this should be much much clearer now :) 
Removed the payloadCorrelationKeys and updated the examples to show how "correlationKey" can be used in all cases.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -192,13 +192,9 @@
...
-        "expression": {
-          "description": "Boolean expression which consists of one or more Event operands and the Boolean operators",
-          "$ref": "#/definitions/expression"
-        },
-        "timeout": {
-          "type": "string",
-          "description": "Time period to wait for incoming events which match the expression (ISO 8601 format)"
+        "eventRefs": {
+          "type" : "array",
+          "description": "References one more more unique event names in the defined workflow events"
```
mswiderski: looks like a typo `one more more` while it should be `one or more`

workflow/spec/spec.md
```diff
@@ -257,24 +257,12 @@ Defines the main structure of serverless workflows:
...
-with each other, and is embedded into the event message.
-Workflow implementations can use this token to map a particular event to a particular workflow instance.
-
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | name | Unique event name | string | yes |
 | source | CloudEvent source | string | yes |
 | type | CloudEvent type | string | yes |
-| correlationToken | Location Path in the event message used to retrieve a token for event correlation | string | no |
+| correlationToken | Context attribute name of the CloudEvent which value is to be used for event correlation | string | no |
```
mswiderski: +1 to this one, makes it clear and easy to follow what it is and how will be used

workflow/spec/schema/serverless-workflow-schema-01.json
tsurdilo: fixed. thanks.

tsurdilo:
Will assume we are good to go on this unless anyone else would like to comment please.

tsurdilo:
I will go ahead and merge as no comments - assume it is ok with everyone. If you have some issues at a later point feel free to open issue and we will address it asap.

Issue #164, opened 2020-02-19
Transition precedence with the retry definition


manuelstein:
Transitions are a great way to specify that the workflow should switch to another state. But I think there's a problem with the retry definition allowing to define transitions.

Currently, transitions can be specified 
- in onError behaviour of (1) workflow runtime errors, (2) delay- (3) event- (4) operation- (5) parallel- (6) switch- (7) subflow- and (8) foreach-state
- in eventactions to evaluate at the end of a sequence or parallel set of actions as part of the event state
- in states to define where next to transition to (delay, operation, parallel, subflow, relay, foreach and switch state) and in switch not only as default but also for single- and- or- and not-choices

and also:
- in retry behaviour of a single action

I'm especially concerned with the latter in concurrent actions. Any parallel state or an event state that uses parallel mode IIUC may contain such an action retry definition, that, when the maximum number of retries has been exceeded, transitions out of the branch into a state that, e.g. halts the workflow. When the retry policy fires, i.e. it says that it is time to transition, the parallel state could no longer wait until completion of all the parallel actions. It is undefined what happens to the ongoing concurrent executions.

Unlike the retry (that is attached to an action) there is no problem with onError, because error definitions are only used for the outcome of the entire state or for overall workflow runtime errors. But retry is tied to an action, which can happen anywhere during a more complex state execution.

I see a several ways to resolve this:

1. specify that an overrun of attempts is an error, i.e. halt or conclude executions within the state to ensure completion and then trigger the state's onError behaviour
2. specify that in case of exceeding maxAttempts, the surrounding state would still conclude (join parallel executions) but that the retry transition overwrites any other evaluation, i.e. a preceding retry transition overrules any other outcome of the state.
3. let the parallel branch enter that state and be generally aware of branching in the workflow

For a clean workflow state handling, I'd prefer **1.** because exceeding retries to me are an error and circumventing the state's onError definition with an in-place retry transition can be a bit confusing.
I'm tempted towards **3.**, i.e. to allow branching/concurrent paths in the workflow. As of now, concurrency can only happen within a state. Even multiple arriving events that correlate to the same workflow instance can not preempt/interrupt the flow but would be buffered and hence may actually never be consumed. At least I think the current positions in this group are not to allow a workflow to take concurrent paths. Branching out and aggregating branches at a later point would be a bit more powerful, but can also make it a lot more complex.

[...]

PR #166, opened 2020-02-21
onError and retry definition updates

tsurdilo:
Clearly defines onError and retry definitions for states.
retry only allowed in Event, Operation, Parallel, and ForEach states. 
Added examples on how to use onError, retry, and combination of both for powerful error handling within states.

### DIFF
--- a/workflow/spec/schema/serverless-workflow-schema-01.json
+++ b/workflow/spec/schema/serverless-workflow-schema-01.json
@@ -145,7 +145,7 @@
           "$ref": "#/definitions/errordatafilter"
         },
         "transition": {
-          "description": "Next transition of the workflow when errors expressed in errorExpression are matched",
+          "description": "Next transition of the workflow when expression is matched",
           "$ref": "#/definitions/transition"
         }
       },
@@ -233,8 +233,8 @@
       "type": "object",
       "description": "Retry Definition",
       "properties": {
-        "errorExpression": {
-          "description": "Expression that matches against error properties",
+        "expression": {
+          "description": "Expression that matches against state data output",
           "$ref": "#/definitions/expression"
         },
         "interval": {
@@ -249,7 +249,7 @@
         }
       },
       "required": [
-        "errorExpression"
+        "expression"
       ]
     },
     "function": {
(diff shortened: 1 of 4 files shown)


tsurdilo:
Related to Issue https://github.com/cncf/wg-serverless/issues/164

tsurdilo:
@manuelstein would you mind taking a quick look please? Thanks.

manuelstein:
Alright, thank you for preparing the PR

1. Workflow runtime error handling ($.onError) is removed as global configuration.
This was not defining default behaviour but handling of a **workflow runtime error**. As this is a global error (the workflow runtime has entirely failed, behaviour undefined), there is no point to make a transition anymore.

2. Retry has been taken out of the action definition and instead can be specified for **event** state, **operation** state, **parallel** state and **foreach** state, because only these state types carry actions. The execution of a remote insecure/third-party action may fail due to temporary errors that are "retry-able", e.g. unavailability of some kind.
The **subflow** state doesn't get a retry behaviour because it may run (and fail) independently with the calling state not waiting on its completion (or failure)

3. correction in errordatafilter description

Only one suggestion: The PR changes the retry definition's *expression* to *errorExpression*, maybe that should remain *expression*.

workflow/spec/schema/serverless-workflow-schema-01.json
```diff
@@ -259,27 +243,23 @@
       "type": "object",
       "description": "Retry Definition",
       "properties": {
-        "expression": {
-          "description": "Boolean expression that matches against the function results. Must be evaluated to true for retry policy to trigger",
+        "errorExpression": {
+          "description": "Expression that matches against error properties",
```
manuelstein: The retry definition specifies an expression to match the state's output against to check whether it needs to be retried. I could imagine that also non-error outcomes may cause retries, e.g. to monitor an external operation. A complex action that retrieves a status may return with "PENDING" and the workflow may want to wait until the returned status is "COMPLETE", which doesn't involve any errors.

tsurdilo:
@manuelstein thanks for comments - updated to rename errorExpression to expression as you mentioned. Also defined that this expression is evaluated against the states data output so that your use case in second comment can be accomplished as well.

manuelstein:
LGTM

tsurdilo:
rebased

PR #174, opened 2020-03-03
New State: Callback State

tsurdilo:
This PR adds a new state: Callback State

While some work performed 
in a serverless workflow can be executed automatically, some decisions must involve manual steps (human decisions for example). 
Callback state allows you to explicitly model manual decision steps during workflow execution.

This state executes a single action (call to a third-party service) where the manual decision needs to be made. The state then waits for a CloudEvent which denotes the completion of this manual decision and contains the decision data in its payload.

Also adding example using callback state


tsurdilo:
rebased an updated image after media dir restructure

tsurdilo:
rebased

workflow/spec/examples.md
```diff
@@ -1846,3 +1847,255 @@ states:
...
+{
+    "id": "customercreditcheck",
+    "version": "1.0",
+    "name": "Customer Credit Check Workflow",
+    "description": "Perform Customer Credit Check",
+    "startsAt": "CheckCredit",
+    "functions": [
+        {
+            "name": "callCreditCheckMicroservice",
+            "resource": "creditCheckResouce",
```
ricardozanini:
```suggestion
            "resource": "creditCheckResource",
```

tsurdilo: fixed (there was one more occurrence of the typo on the yaml definition).

tsurdilo:
Thanks for review @ricardozanini !

tsurdilo:
Regarding the question asked in a meeting if this can be accomplished with 2 states -- yes however for the callback pattern we need a single state (wait state) to be able to correlate back to workflow instance + state after the corresponding event is available (manual task completed). 

We need to support this pattern for 0.1 version. For 0.2 we will enhance it further.

PR #178, opened 2020-03-04
Adding spec roadmap document

tsurdilo:
We really need a roadmap document so can show what has been done and what we are planning to do.
Please review and give comments if anything needs to be added to the "Setup" milestone.

You can see the parsed markdown here: https://github.com/tsurdilo/wg-serverless/blob/addroadmap2/workflow/spec/roadmap.md

### DIFF
--- a/workflow/spec/roadmap.md
+++ b/workflow/spec/roadmap.md
@@ -1,0 +1,45 @@
+# Roadmap
+
+The Serverless Workflow Roadmap.
+
+_Note: Items in tables for each milestone do not imply an order of implementation._
+
+_Note: Milestone entries include the most notable updates only. For list of all commits see [link](https://github.com/cncf/wg-serverless/commits/master)_
+
+_Status description:_
+
+| Completed | In Progress | In Planning |
+| --- | --- |  --- | --- |
+|  <center>✔</center> | <center>✏️</center> | <center>🚩</center> |
+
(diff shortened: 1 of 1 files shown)


tsurdilo:
@duglin Would you mind taking a look please? Thanks.

duglin:
Can't comment on the individual work items since you guys are in a better position to know that, but I like the overall flow and agree it's needed!  My only comment is more procedural... do you want to create PRs each time you want to edit this table?  Since you point to PRs in it, you might be doing it very very often and that could get annoying. It might be easier if you just listed high-level goals for each milestone/release in this doc and then point people to a github issue query that showed all of the issues with the specified milestone or label. 

Not saying CE's was great, but our doc listed pretty high-level things:  https://github.com/cloudevents/spec/blob/master/roadmap.md  and then we used labels to tag issues for each release. We then updated the roadmap at the end of each release based on what really happened.  :-)

tsurdilo:
@duglin I thought about this (and looked at the CE readme as well). I think at this point where we are still small and trying to attract community as well as show off our work to TOC the readme is ok to show a number of the bigger updates/changes we have added to the specification.

There is only 1 link to a PR, other links point to the spec / use-cases / examples docs. 

I think in the near future hopefully we can switch to the same style as CE.

tsurdilo:
rebased and added on-hold icon

PR #179, opened 2020-03-05
Small doc restructure and update

tsurdilo:
* Adds a description image of the top-level workflow markup:

Screen Shot 2020-03-05 at 12 03 16 PM

* Presents overview of states in table format rather than simple list:

Screen Shot 2020-03-05 at 12 03 30 PM

* restructures definitions to make more sense

* rewords some introduction text to be more clear


tsurdilo:
rebased and updated image after media dir restructure

PR #180, opened 2020-03-05
Update main readme doc

tsurdilo:
* Updates the main description of the specification to make it more clear. Also aligns it with the introduction section of the main spec document.

* Adds meeting time information

* Adds meeting minutes information

### DIFF
--- a/workflow/spec/readme.md
+++ b/workflow/spec/readme.md
@@ -1,23 +1,25 @@
-# Serverless Workflow
-
 <p align="center">
 <img src="media/sw-logo.png" alt="Serverless Workflow Specification"/>
 </p>
 
-Serverless applications are becoming increasingly complex. Nowdays they have to coordinate, manage, and define
-the execution order (steps) for countless functions triggered by as many events.
+Serverless applications are becoming increasingly complex.
+Often they contain complex business logic to define, coordinate, and manage the execution
+order of a large amount of serverless functions and events that can trigger those functions.
 
-Workflows have become key components of serverless applications as they excel at orchestration and coordination
-of their functional flow. 
+Workflows have become a key component of serverless applications as they provide
+clear separation between business and orchestration logic.
 
-The goal of the Serverless Workflow sub-group is two-fold:
+Serverless Workflow is a specification that strandardizes the model for workflows
+responsible for orchestrating event-driven serverless applications.
 
-- Come up with a standard language for users to specify their serverless application workflow, as well as
-- Provide a Cloud Provider agnostic API to integrate Workflows into the ecosystem.
+Main goals of the specification include:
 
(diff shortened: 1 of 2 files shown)


workflow/spec/readme.md
```diff
@@ -1,23 +1,25 @@
...
+Often they contain complex business logic to define, coordinate, and manage the execution
+order of a large amount of serverless functions and events that can trigger those functions.
 
-Workflows have become key components of serverless applications as they excel at orchestration and coordination
-of their functional flow. 
+Workflows have become a key component of serverless applications as they provide
+clear separation between business and orchestration logic.
 
-The goal of the Serverless Workflow sub-group is two-fold:
+Serverless Workflow is a specification that strandardizes the model for workflows
```
duglin: the standards geek in me is coming out.... I'd prefer "defined" instead of "standardizes" since to me a "standard" is more than just a spec - and this is just a spec right now.  But if you keep it as-is, you misspelled it  :-)   Remove the first "r".

workflow/spec/readme.md
```diff
@@ -1,23 +1,25 @@
...
 
-- Come up with a standard language for users to specify their serverless application workflow, as well as
-- Provide a Cloud Provider agnostic API to integrate Workflows into the ecosystem.
+Main goals of the specification include:
 
-This will facilitate the portability of serverless applications across different vendor platforms.
+- To facilitate Serverless Workflow portability across different vendor platforms
+- To be completely vendor neutral
+- To support both stateless and stateful Serverless Workflow orchestration
+- To define a light-weight, human-readable, and embeddable Serverless Workflow model
```
duglin: Not really objecting, but what do you mean by "human-readable"? My mind jumps to something closer to "natural language" than json. Or is this just your way of saying "it's not some binary format"? 

And what does "embeddable" mean?  What do you envision it being embedded into ?

duglin:
Overall, it seems good to me. Just had some minor questions that apply to both docs, but only left them on the first since the text is duplicated.

workflow/spec/readme.md
tsurdilo: updated

workflow/spec/readme.md
tsurdilo: "human readable" was attempt to say that unlike other workflow models that completely reply on visualization tooling, serverless workflow model can be read and understood by humans without need of tooling/visualization.

"embeddable" means that especially since we support yaml format as well, serverless workflow "files" can be either stand-alone or embedded as part of things like kubernetes deployments etc. 

After thinking about it just replaced with "powerful" as otherwise we should probably define a detailed description of what exactly we need by the other terms..i can do that with separate pr in future.

PR #183, opened 2020-03-09
Remove the 'spec-' prefix from specification documents

tsurdilo:
Just removes the "spec-" prefix from specification docs and updates links


tsurdilo:
does not change spec, just a cleanup pr

PR #184, opened 2020-03-09
Adding Fako Menge to list of contributors

tsurdilo:
..and small reformatting...


tsurdilo:
there was no objections during meeting. adding

PR #203, opened 2020-03-16
Mocking Workflow Data Section

tsurdilo:
This pr adds the "Mocking Workflow Data" section and introduces a way to mock workflow data
during testing. 

It also defined the data-mock json schema which can be used for data mock files.

### DIFF
--- a/workflow/spec/schema/data-mock.json
+++ b/workflow/spec/schema/data-mock.json
@@ -1,0 +1,78 @@
+{
+    "$id": "https://wg-serverless.org/data-mock.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "Serverless Workflow specification Data Mock schema. Used to define static data to be used during workflow execution.",
+    "type": "object",
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Workflow name",
+            "minLength": 1
+        },
+        "description": {
+            "type": "string",
+            "description": "Workflow description"
+        },
+        "eventsData": {
+            "type": "array",
+            "description": "Defins data mocks for events data",
+            "items": {
+                "type": "object",
+                "$ref": "#/definitions/eventDataDef"
+            }
+        },
+        "actionsResult": {
+            "type": "array",
+            "description": "Define data mocks for actions result",
+            "items": {
+                "type": "object",
+                "$ref": "#/definitions/actionResultDef"
+            }
+        },
+        "workflowResult": {
+            "type": "object",
+            "description": "Workflow data result"
+        }
+    },
+    "required": [],
+    "definitions": {
+        "eventDataDef": {
+            "type": "object",
+            "description": "Mock event data",
+            "properties": {
+                "eventDefName": {
+                    "type": "string",
+                    "desription": "Unique name of defined event in the workflow definition",
+                    "minLength": 1
+                },
+                "data": {
+                    "type": "object",
+                    "description": "Mock event data"
+                }
+            },
+            "required": [
+                "eventDefName",
+                "data"
+            ]
+        },
+        "actionResultDef": {
+            "type": "object",
+            "description": "Mock action result",
+            "properties": {
+                "actionDefName": {
+                    "type": "string",
+                    "desription": "Unique name of defined action in the workflow definition",
+                    "minLength": 1
+                },
+                "result": {
+                    "type": "object",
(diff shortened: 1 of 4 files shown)


workflow/spec/schema/data-mock.json
```diff
@@ -0,0 +1,78 @@
...
+            "description": "Workflow name",
+            "minLength": 1
+        },
+        "description": {
+            "type": "string",
+            "description": "Workflow description"
+        },
+        "eventsData": {
+            "type": "array",
+            "description": "Defins data mocks for events data",
```
falko: Typo: Defins => Defines

workflow/spec/spec.md
```diff
@@ -253,6 +255,11 @@ As mentioned, implementation compliance is based on the workflow definition lang
           "items": {
             "type": "object"
           }
+        },
+        "dataMock": {
+          "type": "string",
```
falko: should this by of type URL?

workflow/spec/spec.md
```diff
@@ -3516,6 +3523,83 @@ the "afterFunctionErrorState" state. In case of any other errors, they will be r
 If those errors occur after two max 2 retries, the onError definition states that workflow should transition to the
 "afterAnyOtherErrorState" state.
 
+## Mocking Workflow Data
+
+When testing your serverless workflows, you might not be ready to call some or all services invoked during workflow execution.
+You also might want to use pre-defined event data for some or all events that trigger actions execution or workflow intantiation.
+In certain cases you might even want to pre-define the enire workflow data result itself.
+
+Serverless Workflow provides the [data mock JSON Schema](schema/data-mock.json) which you can use to define reusable workflow data mocks for testing.
```
falko: Specification language should probably be more formal and not address the reader directly with `you`. How do we handle this in the rest of the document?

workflow/spec/spec.md
tsurdilo: some impls might want a relative path or classpath definition like for example

"./somemockfile.json"

so no need to do that

workflow/spec/spec.md
tsurdilo: There are lots of references to "you" "your". Whats suggested change?

tsurdilo:
@falko it is the intent to define a TCK as part of **this** specification and not a separate effort.

tsurdilo:
> On the one hand, this goes beyond the scope of the specification. On the other hand, it doesn't go far enough as to defining complete test cases. This could be a foundation for a TCK, which could, however, be a separate project as demonstrated with DMN.

I understand the concern but don't think this is necessarily the case. The specification already defines in the workflow data section that:
* Actions have a result which is merged into state data
* Events have a payload (data) which is merged into the state data
* Workflow returns a result 

So I don't really see how this goes beyond the scope of the specification. It simply defines that which is merged from "outside" sources can be mocked.

workflow/spec/spec.md
falko: Replace second-person pronouns using the [classic tricks of technical writing](https://semo.edu/pdf/Writing_handout_Avoiding_Second_Person.pdf).

workflow/spec/schema/data-mock.json
tsurdilo: fixed. thanks!

workflow/spec/spec.md
tsurdilo: Ok. I will create a separate PR using these suggestions for all of the spec docs. Thanks!

falko:
It all comes back to the question of how big or small of a specification this should become. What do the others think?

tsurdilo:
> It all comes back to the question of how big or small of a specification this should become.

I think this question assumes there are some artificial size boundaries on the specification, which do not exist. Big/small compared to what? There is no intent to alter the specification growth based on size or possible comparisons to other specifications.

tsurdilo:
Removed the "dataMock" property from workflow definition. This pr only defines a schema that can be used during testing to mock data results of service invocations and used event payloads.

workflow/spec/spec.md
tsurdilo: see https://github.com/cncf/wg-serverless/pull/205. it was suggested by @lucperkins not to apply these changes. thanks!

tsurdilo:
rebased

tsurdilo:
might revisit this after v0.1

PR #218, opened 2020-03-30
Replace Mauricio with Falko as owner representing Camunda

falko:
As suggested by @tsurdilo, @salaboy and I decided that it would be more suitable for the workflow project to switch roles.


tsurdilo:
lgtm

tsurdilo:
was approved in serverless workflow team meet April 6. Congrats Falko!

PR #225, opened 2020-04-07
Rename Relay to Inject state

tsurdilo:
Renames Relay state to Inject state so it's name makes more sense.

After getting some info from community, one of the bigger complaints was that noone understood what the name "Relay" means in relay state. It is a state that injects static data into the state data input. So from that changing:

Relay State / inject property 
to
Inject State / data property

to clearly describe that this is a state that "injects" "data".

### DIFF
--- a/workflow/spec/examples.md
+++ b/workflow/spec/examples.md
@@ -20,7 +20,7 @@
 
 #### Description
 
-This example uses two relay states. The "Hello" state statically injects the following JSON into its data input:
+This example uses two inject states. The "Hello" state statically injects the following JSON into its data input:
 
 ```json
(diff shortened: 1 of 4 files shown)


cathyhongzhang:
Is this "injected data to the state" in addition to data passed to this state 
1. from the previous state output
2. from the previous state's functions' output
3. from external events if this state is an event state

tsurdilo:
@cathyhongzhang Hi :) yes the inject state has state data input like all the others. 
It injects additional static data into its state data. This data is merged with its existing data input.

cathyhongzhang:
LTGM

PR #226, opened 2020-04-07
Updating Switch State

tsurdilo:
Updates Switch state:
* renames "choices" to "conditions"
* Removes the unnecessary "and "not" "or" choices and replaces them with condition definitions
* Adds more values as well as "Custom" value to condition operator
* Updates all examples

Reason: Our Switch state is very important state in spec however was way too complex to use. It acts as an exclusive gateway so the docs and its definition have been updated to clearly state its purpose. The different choices were not needed and just made things more complicated than should be, so they are removed.

### DIFF
--- a/workflow/spec/examples.md
+++ b/workflow/spec/examples.md
@@ -694,7 +694,7 @@ This example shows off the switch state and the subflow state. The workflow is s
     }
 ```
 
-We use the switch state with two choices to determine if the application should be made based on the applicants age.
+We use the switch state with two conditions to determine if the application should be made based on the applicants age.
 If the applicants age is over 18 we start the application (subflow state). Otherwise the workflow notifies the
  applicant of the rejection.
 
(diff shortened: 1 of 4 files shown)


wanghq:
Why renaming the choices to conditions? I feel those are actually choices which contain condition and actions (goto).  Or if you don't like the `choices`, is `cases` an ok name to follow the programming lang convention like https://en.wikipedia.org/wiki/Switch_statement ?

tsurdilo:
@wanghq I understand your point and in general you are correct. 
Traditionally the structure as you describe is
"switch state includes 1 or more choices/cases, each choice/case contains a condition and action/transition."

We simplify this and just have conditions, which include a transition definition. So for what we define can be described as:
"switch state includes 1 or more conditions which can trigger transitions(actions)".

IMO this reduces the amount of json to define the same functionality. 
WDYT?

tsurdilo:
rebased

PR #227, opened 2020-04-07
Update state definitions table

tsurdilo:
Updates the state definitions table:
* adds new column "Can be workflow start state"
* adds new column "Can be workflow end state"

This is to clearly show that switch state cannot be a workflow end state up-front in this table.


tsurdilo:
rebased

PR #229, opened 2020-04-09
Parallel State - waitForCompletion update

tsurdilo:
Moves the waitForCompletion property from branch to parallel state.
The parallel state itself needs to determine how to "join" branches, and not each branch itself.

### DIFF
--- a/workflow/spec/examples.md
+++ b/workflow/spec/examples.md
@@ -558,9 +558,8 @@ states:
 #### Description
 
 This example uses a parallel state to execute two branches (simple wait states) at the same time.
-Note that the waitForCompletion flag is set to "false" so as soon as the "ShortDelay" delay state finishes,
-the workflow complete execution. If waitForCompletion was set to true, the workflow would complete after both
-of the branches are done.
+The waitForCompletion flag is set to false, which means the parallel state has to wait for both branches
+to finish execution before it can transition (end workflow execution in this case as it is an end state).
 
 #### Workflow Definition
 
@@ -585,6 +584,7 @@ of the branches are done.
      "start": {
        "kind": "DEFAULT"
      },
+     "waitForCompletion": true,
      "branches": [
         {
           "name": "Branch1",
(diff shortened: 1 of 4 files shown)


falko:
LGTM

workflow/spec/examples.md
```diff
@@ -558,9 +558,8 @@ states:
 #### Description
 
 This example uses a parallel state to execute two branches (simple wait states) at the same time.
-Note that the waitForCompletion flag is set to "false" so as soon as the "ShortDelay" delay state finishes,
-the workflow complete execution. If waitForCompletion was set to true, the workflow would complete after both
-of the branches are done.
+The waitForCompletion flag is set to false, which means the parallel state has to wait for both branches
```
wanghq, line 561: the false should be true?

tsurdilo:
> This allows to "wait for all" or "wait for one". What if someone asks for "wait for N branches"?

Good point. I will update and add other options as well

tsurdilo:
@wanghq Updated the pr, waitForCompletion is now an enum "completionType" with values:
"AND", "XOR", "N_OF_M". Default is "AND". 
Also added "n" property to define the 'N' in 'N_OF_M. 

I think this covers the case you mentioned in comment.

@falko do you mind to review again after these updates? Thanks.

wanghq:
LGTM, thanks.

falko:
Which workflow patterns or BPMN gateway types is `N_OF_M` targetting?

falko:
It sounds like a completion condition on a BPMN multi-instance activity. An inclusive join can't be achieved with this unless n is somehow dynamic.

tsurdilo:
@falko I don't think this is targeting any bpmn patterns. It just describes a completion where n number of branches have to complete. Many other models might have a similar pattern of completion, we are not copying anything from bpmn.

Regarding **An inclusive join can't be achieved with this unless n is somehow dynamic.** - yes this will be possible very soon. You will be able to say "n": "$.xyz" where xyz is a property of state data. Since this involves alot of properties of the current model, I will add that as part of a separate pr.

tsurdilo:
rebased

PR #239, opened 2020-05-08
Switch State - add event-based conditions

tsurdilo:
* Changes "conditions" to two types: "dataConditions" and "eventConditions"
* One of the two types of conditions are allowed to be used within a single switch state
* "dataConditions" acts exactly same as before - transitions are made based on state data
* "eventConditions" can be used to make event based transitions. In this case Switch state acts as as a wait state, waiting for the condition events to arrive. When one arrives its associated transition is made. If none arrive in a timeout period a default transition is taken.
* Adds event-based switch state example
* Updates all examples to reflect the made changes.
* Small fixes to properties in tables (spec doc)

Reason: Currently our switch state is only data based. It is important to be able to make workflow transitions as well based on events.


tsurdilo:
rebased

tsurdilo:
assume noone has any objectsions :) merging and feel free to add issues/comments whenever you might have time to review. Thanks!

PR #258, opened 2021-12-06
explain how we deal with sub-projects

duglin:
Signed-off-by: Doug Davis <dug@us.ibm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,17 +7,17 @@ work the following artifacts were produced:
 * [A Serverless Overview Whitepaper](#serverless-overview-whitepaper)
 * [Landscape](#landscape)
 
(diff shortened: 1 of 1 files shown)


markpeek:
Should this change talk about, as the sub-group matures, the route of becoming a standalone project (sandbox, etc.) via the CNCF TOC guidelines/approval?

duglin:
@markpeek does the maturity level of the project change whether it's part of our WG? I was assuming it didn't. Are you looking for something in this `...each sub-project is independent and defines their own processes, governance model and work.` ? Perhaps add "CNCF status" ??

duglin:
Approved on the 1/13 call

PR #259, opened 2023-01-28
Update README.md

Jay1305:
Interactive url looks like changed from https://landscape.cncf.io/format=serverless to https://landscape.cncf.io/serverless. The easy to remember url s.cncf.io also needs to change to point to correct url. But probably only owner can change it.

Signed-off-by: Jay1305 <31636584+Jay1305@users.noreply.github.com>


duglin:
Thanks for noticing!

tsurdilo (approved):

Issue #261, opened 2024-02-06
Formalize the charter for the Serverless WG


jasonsmithio:
We have a "charter-esque" like listing [here](https://github.com/cncf/wg-serverless/tree/d3f38320f1bb85b45b510bb9c33d3478522339a7?tab=readme-ov-file#goals) but I think we should formalize not just out goals but our operations, how we vote, etc. 

I am happy to take the charge. Creating this issue for tracking purposes.

jasonsmithio:
PR for first pass #262

PR #262, opened 2024-06-19
Charter per Issue #261

jasonsmithio:
First pass of a charter per issue #261

### DIFF
--- a/proposals/charter.md
+++ b/proposals/charter.md
@@ -1,0 +1,63 @@
+---
+title: CNCF Serverless Working Group Charter
+url: 'proposals/'
+---
+
+## Introduction
+The Serverless Working Group, is a small group formed by passionate serverless practitioners who are looking to expand the serverless landscape and provide guidance in how it interacts with the larger cloud native landscape.
+
+## Background
+
+While Kubernetes adoption is undeniably on the rise, many plaftorm engineers are experiencing friction in adoption with their developers. Kubernetes is a great platform but isn't the right level of abstraction for developers. This group will provide practices to improve infrastructure abstraction from developers to improve Kubernetes adoption. 
+
+## Goals
+
+This workgroup's objectives are as follows:
+
+- Provide clarity for ourselves and consumers w.r.t. what this topic is all about:
+  - Define common terminology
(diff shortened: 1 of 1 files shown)


jasonsmithio:
signed work

duglin:
@jasonsmithio I'm wondering if it's worth a separate "charter" doc when it says the same thing as the main README. Would it make more sense to just edit the README to make it clear that the top section is basically our "charter" or "goals"? Then the following section (non-goals) would flow nicely from that.

We probably need to clean-up some stuff in there anyway - like add xRegistry to the list of projects.

jasonsmithio:
This is just meant to be a starter but I am happy to update it with more details. 

I do have a few questions thought. Is everything in the [Serverless Landscape](https://landscape.cncf.io/?group=serverless) under our purview or just CloudEvents, Serverless Workflow and xRegistry? I also think there may be value in adding a simple glossary in the charter to describe key words such as "eventing" and such. 

I am going to us some of the other WG charters for more ideas.

duglin:
We're definitely not just CE, Workflow and xReg, qnd we basically decide for ourselves what our next project will be. The problem is that a lot of the terms used (even "serverless" itself) is vague enough that it could mean anything. E.g. CloudEvents isn't serverless specific.  Net: we're kind of winging it. :-) 

re: glossary... maybe. I think the serverless whitepaper already has some of those - but be prepared for rathole discussions. However, it would help me if we knew what the pain point is that we're trying to address with this PR. I don't disagree that some stuff is stale and could use some clean-up, but I doubt there's much of a desire to spend too much time working on new docs "just because". That's why I'd prefer smaller, targeted, PRs that update the current text so it doesn't look old/stale rather than creation of new content unless there's a concrete need to do so.

As an example... in general the idea of a "charter" is fine, but when its content is already there in the README, the effort to convert it into a document as formal as a "charter" implies a lot of work and thinking because of the seriousness of the term. Combine that with the fact that we've gone years w/o a "charter" w/o any complaints or issues, it kind of brings up the question of "why now?" or "who really cares or needs it?" 

As you might have figure out, I'm not a fan of bureaucracy - but that's just me

PR #263, opened 2025-03-14
Updated the existing drawback and made it more comprehensive

azarboon:
This PR updates an existing drawback in the Serverless Overview whitepaper (Functions-as-a-Service section) and makes it more comprehensive. It highlights the overuse of serverless functions as an antipattern. Specifically, it addresses cases where functions are used instead of service-native integrations and when functions trigger other functions. These common antipatterns can introduce unnecessary complexity, operational overhead, performance inefficiencies, and challenges in testing, debugging, and observability.

### DIFF
--- a/whitepapers/serverless-overview/README.md
+++ b/whitepapers/serverless-overview/README.md
@@ -386,7 +386,7 @@ Drawbacks include less comprehensive and stable documentation, samples, tools, a
 
 * Due to the on-demand structure, the "cold start" aspect on some serverless runtimes could be a performance issue if the runtime removes all instances of a function when idle.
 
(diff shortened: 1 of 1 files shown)


clemensv:
It's not clear to me what the motivation for this contribution is. It seems to be a fairly random edit that I don't see as adding substantial value. I also don't see the existing text as being confusing. 

It's a platitude that using a thing for scenarios that it's not well suited for is problematic.

jskeet:
I agree with @clemensv that this isn't needed - particularly not on a whitepaper from several years ago.

duglin:
Discussed on the 3/20 call and the group agreed to not make this change for the reasons stated in the previous comments

PR #264, opened 2025-08-06
Remove unusual line terminator from README

HayatoW:
This pull request makes a small improvement to the documentation in `whitepapers/serverless-overview/README.md`. The change removes an unnecessary special character from the drawbacks section, improving readability.


duglin:
that's really weird. Thanks for noticing and for the PR

